PR #29, opened 2017-06-16
[WIP] Documentation

vitvakatu:
- [x] Provide top-level documentation for every struct.
- [x] Provide documentation for every entity, field and method.
- [ ] Provide in-documentation examples.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,22 @@
+#![warn(missing_docs)]
+//! # Motivation and Goals
+//!
+//! 1. Ergonomics is first priority. Being able to prototype quickly and code
+//! intuitively is more important than capturing all the 3D features. We already
+//! have a solid foundation with [gfx-rs](https://github.com/gfx-rs/gfx), so
+//! let's make some use of it by providing a nice higher-level abstraction.
+//! 2. Follow "Three.JS" - this is simpler than coming up with a brand new API
+//! (like [kiss3d](https://github.com/sebcrozet/kiss3d)), yet instantly familiar
+//! to a large portion of Web develper. Some deviations from the original API are imminent.
+//! 3. Explore the capabilities of Rust to mimic deep object-oriented nature of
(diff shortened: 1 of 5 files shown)


src/scene.rs
```diff
@@ -204,6 +237,9 @@ impl AmbientLight {
     }
 }
 
+/// The light source that illuminates all objects equally from a given direction,
+/// like an area light of infinite size and infinite distance from the scene;
```
kvark: I don't think "infinite size" needs to be here

kvark:
The in-documentation examples is something I'm not 100% sure about, since even if you make them CI-tested, they still add up the maintenance cost, especially now that the library is still evolving rapidly.

src/scene.rs
vitvakatu: So, only "The light source that illuminates all objects equally from a given direction"?

kvark: that would be fine, yeah

src/lib.rs
```diff
@@ -1,3 +1,22 @@
+#![warn(missing_docs)]
+//! # Motivation and Goals
+//!
+//! 1. Ergonomics is first priority. Being able to prototype quickly and code
```
kvark: I'm not keen on the idea of duplicating this both here and in the Readme...

vitvakatu: Oh, yeah, I forgot to say that. I need help to write good module-level documentation :D

kvark: I think not having any is somewhat better than duplication, so it's fine to remove that.

vitvakatu: But we have `#![warn(missing_docs)]`. Should I remove that? (I've documented everything)

kvark: No, just have a one-liner for the module documentation

kvark:
See, now it's trying to test your in-doc tests. At this point, we need to either implement OsMesa support for headless rendering, or not have those in-doc examples at all.

PR #32, opened 2017-06-18
Update graph in sync

vitvakatu:
I think it was mistake not to call `update_graph` in `sync()`.
Without this, `NodeInfo` doesn't contain *actual* info and `sync()` panics on `scene_id` assertion.


kvark:
Thanks!
bors r+

PR #33, opened 2017-06-19
New input API

### LINKED ISSUE #8: Input format
Three.js doesn't appear to provide a single structure for all the input per frame.
Main problem is - some use cases need an impulse mode (has "X" just been hit?), some need a modal one (is "x" currently pressed?). We night end up doing both.

kvark:
Closes #8

r? @vitvakatu

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,0 +1,171 @@
+#![allow(missing_docs)] //TODO
+
+use glutin::{MouseButton, ElementState, VirtualKeyCode as Key};
+use mint;
+
+use std::collections::HashSet;
+use std::time;
+
+
+pub type TimerDuration = f32;
+
+#[allow(dead_code)]
+struct InputState {
+    time_moment: time::Instant,
+    is_focused: bool,
+    keys_pressed: HashSet<Key>,
+    mouse_pressed: HashSet<MouseButton>,
+    mouse_pos: mint::Point2<f32>,
+}
+
+#[allow(dead_code)]
+struct InputDiff {
+    time_delta: TimerDuration,
+    keys_hit: Vec<Key>,
+    mouse_moves: Vec<mint::Vector2<f32>>,
+    mouse_hit: Vec<MouseButton>,
+}
+
+pub struct Input(InputState, InputDiff);
+
+impl Input {
+    pub fn new() -> Self {
+        let state = InputState {
+            time_moment: time::Instant::now(),
+            is_focused: true,
+            keys_pressed: HashSet::new(),
+            mouse_pressed: HashSet::new(),
+            mouse_pos: [0.0; 2].into(),
+        };
+        let diff = InputDiff {
+            time_delta: 0.0,
+            keys_hit: Vec::new(),
+            mouse_moves: Vec::new(),
+            mouse_hit: Vec::new(),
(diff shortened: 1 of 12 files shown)


src/input.rs
```diff
@@ -0,0 +1,172 @@
...
+    is_focused: bool,
+    keys_pressed: HashSet<Key>,
+    mouse_pressed: HashSet<MouseButton>,
+    mouse_pos: mint::Point2<f32>,
+}
+
+#[allow(dead_code)]
+struct InputDiff {
+    time_delta: TimerDuration,
+    //time_delta_focused: time::Duration,
```
vitvakatu: Do we need this?

src/input.rs
```diff
@@ -0,0 +1,172 @@
...
+        });
+        self.0.mouse_pos = pos;
+    }
+}
+
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Timer {
+    start: time::Instant,
+    //focus_only: bool,
```
vitvakatu: And this?

src/input.rs
```diff
@@ -0,0 +1,172 @@
...
+    Mouse(MouseButton),
+}
+
+pub const KEY_ESCAPE: Button = Button::Key(Key::Escape);
+pub const KEY_SPACE: Button = Button::Key(Key::Space);
+
+impl Button {
+    pub fn count_hits(&self, input: &Input) -> u8 {
+        match *self {
+            Button::Key(button) => input.1.keys_hit.iter().filter(|&&key| key == button).take(0xFF).count() as u8,
```
vitvakatu: Any reason to use hex number instead of decimal?
Also, maybe we should use constant for this?

src/input.rs
```diff
@@ -0,0 +1,172 @@
...
+
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct KeyAxis {
+    pub neg: Key,
+    pub pos: Key,
+}
+
+pub const AXIS_LEFT_RIGHT: KeyAxis = KeyAxis{ neg: Key::Left, pos: Key::Right };
+pub const AXIS_DOWN_UP: KeyAxis = KeyAxis{ neg: Key::Down, pos: Key::Up };
```
vitvakatu: It seems irregular for me. Maybe `AXIS_UP_DOWN`? People usually look from the top to the down.

src/input.rs
```diff
@@ -0,0 +1,172 @@
+#![allow(missing_docs)] //TODO
```
vitvakatu: I'll try to fix it asap.

kvark: thanks, and no rush!

src/input.rs
kvark: removed now

src/input.rs
kvark: fixed

src/input.rs
kvark: The axis go from DOWN (negative) to UP (positive), hence the name `AXIS_DOWN_UP`, which is consistent with `AXIS_LEFT_RIGHT`.

kvark:
Comments addressed, unleashing bors.
bors r+

PR #36, opened 2017-06-21
Cube-ception (group) demo

kvark:
Runs quite slow on my machine with 6 levels, presumably because we don't cache the current pipeline, so it's a good testcase.


kvark:
Ok, caching the current PSO is not the problem, since gfx-rs GL backend does that already.

PR #42, opened 2017-06-24
Check window sizes before updating the view

shmutalov:
Fixes crashes on Windows machines when trying to minimize application


vitvakatu:
I don't see the way it can fix this issue.

shmutalov:
Issue is not about freezing OS. It is not related with that bug.

kvark:
no harm accepting this now
thanks!
bors r+

PR #43, opened 2017-06-25
[WIO] Ray casting

kvark:


alteous:
Has there been any progress on this PR?

kvark:
@alteous there is a lot to do here, not much progress since what is published here (distracted by other shiny things).

Rhuagh:
There is now a BVH in collision-rs that has support for raycasting. It's not super fast yet, but I have some ideas for that.

kvark:
@Rhuagh I pushed all of the local code here now. It's hacky and not production ready.
Please feel free to take over, if find this useful.

PR #51, opened 2017-07-12
Add mouse wheel input handling

### LINKED ISSUE #50: Request: expose mouse wheel events
Would it be possible to expose mouse wheel events? My rust skills aren't quite up to it. The mouse wheel is an obvious way to control zoom.

vitvakatu:
Fixes #50

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -73,8 +73,8 @@ impl Input {
         self.0.mouse_pos
     }
 
-    pub fn get_mouse_wheel_movements(&self) -> Vec<f32> {
-        self.1.mouse_wheel.clone()
+    pub fn get_mouse_wheel_movements(&self) -> &[f32] {
+        &self.1.mouse_wheel[..]
     }
 
     pub fn get_mouse_wheel(&self) -> f32 {
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,11 +1,12 @@
 #![allow(missing_docs)] //TODO
 
-use glutin::{MouseButton, ElementState, VirtualKeyCode as Key};
+use glutin::{MouseButton, MouseScrollDelta, ElementState, VirtualKeyCode as Key};
 use mint;
 
 use std::collections::HashSet;
 use std::time;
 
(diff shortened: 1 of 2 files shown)


src/input.rs
```diff
@@ -1,11 +1,12 @@
 #![allow(missing_docs)] //TODO
 
-use glutin::{MouseButton, ElementState, VirtualKeyCode as Key};
+use glutin::{MouseButton, MouseScrollDelta, ElementState, VirtualKeyCode as Key};
 use mint;
 
 use std::collections::HashSet;
 use std::time;
 
+const PIXELS_PER_LINE: f32 = 38.0;
```
vitvakatu: I'm not sure is it correct, I just looked into `servo`.

src/input.rs
```diff
@@ -101,6 +108,13 @@ impl Input {
         });
         self.0.mouse_pos = pos;
     }
+
+    pub fn mouse_wheel(&mut self, delta: MouseScrollDelta) {
+        self.1.mouse_wheel += match delta {
```
kvark: The reason I don't accumulate moues offsets in the same way is because the action user might want to do based on the moves is not necessarily cumulative. For example, consider that the user wants to play a sound on each scroll over a threshold. If you just sum up the events, and they end up being near zero, the user will not register the threshold.

TL;DR: I think we should keep a list of the wheel scrolls, and have a method that returns the total

vitvakatu: I'm not sure is it possible to make two different mouse wheel moves during one frame...

kvark: Consider frame loop and event loop to be independent

vitvakatu: Oh, right. I'll change the behavior

src/input.rs
```diff
@@ -73,8 +73,12 @@ impl Input {
         self.0.mouse_pos
     }
 
+    pub fn get_mouse_wheel_movements(&self) -> Vec<f32> {
```
kvark: similarly, please return `&[]`

PR #52, opened 2017-07-13
Expose mouse_moves

### LINKED ISSUE #49: Expose mouse moves from InputDiff
There is no easy way to get mouse moves (e.g. delta of position since last update) atm.
However, we have all capabilities to easily do that by adding new method to `Input` impl.
cc @shmutalov

vitvakatu:
Fixes #49

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -81,6 +81,19 @@ impl Input {
         self.1.mouse_wheel.iter().sum()
     }
 
+    pub fn get_mouse_movements(&self) -> &[mint::Vector2<f32>] {
+        &self.1.mouse_moves[..]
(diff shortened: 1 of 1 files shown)


src/input.rs
```diff
@@ -70,6 +70,19 @@ impl Input {
         self.0.mouse_pos
     }
 
+    pub fn get_mouse_movements(&self) -> Vec<mint::Vector2<f32>> {
```
kvark: should return `&[]` without cloning

kvark:
Please rebase and squash

vitvakatu:
@kvark ping

PR #54, opened 2017-07-14
Builder patterns

vitvakatu:
Builder patterns for `OrbitControls` and `Window`.


kvark:
Thank you! I think we can make it nicer, but this is a good step forward.

PR #61, opened 2017-07-28
Add PBR material type

alteous:
This PR implements a physically-based rendering pipeline based on the [glTF reference renderer](https://github.com/KhronosGroup/glTF-WebGL-PBR).

There is one example provided, namely `pbr`, which should output the following.

render


data/shaders/pbr_ps.glsl
```diff
@@ -0,0 +1,211 @@
...
+float ggx(PbrInfo pbr)
+{
+    float roughness_sq = pbr.alpha_roughness * pbr.alpha_roughness;
+    float f = (pbr.ndoth * roughness_sq - pbr.ndoth) * pbr.ndoth + 1.0;
+    return roughness_sq / (PI * f * f);
+}
+
+bool available(int flag)
+{
+    return (u_PbrFlags & flag) == flag;
```
alteous: This probably isn't the best strategy for good parallel performance but I'm not sure what else to do.

examples/pbr.rs
```diff
@@ -0,0 +1,160 @@
+
+extern crate gltf;
+extern crate three;
+
+fn load_mesh(mesh: gltf::mesh::Mesh, factory: &mut three::Factory) -> three::Mesh {
```
alteous: Native glTF support 'coming soon'.

src/factory.rs
```diff
@@ -319,6 +367,34 @@ impl Factory {
...
+ 
+    /// Create new `Sampler`.
+    pub fn sampler(
+        &mut self,
+        filter_method: FilterMethod,
+        horizontal_wrap_mode: WrapMode,
+        vertical_wrap_mode: WrapMode,
+    ) -> Sampler {
+        use gfx::texture::Lod;
+        let info = gfx::texture::SamplerInfo {
```
alteous: The `Lod` values were copied from [`gfx::texture::SamplerInfo::new()`](https://docs.rs/gfx_core/0.7.1/src/gfx_core/texture.rs.html#455-468).

src/render.rs
```diff
@@ -475,37 +534,161 @@ impl Renderer {
...
-                vbuf: gpu_data.vertices.clone(),
-                cb_locals: gpu_data.constants.clone(),
-                cb_lights: self.light_buf.clone(),
-                cb_globals: self.const_buf.clone(),
-                tex_map: map.unwrap_or(&self.map_default).to_param(),
-                shadow_map0: (shadow0.clone(), shadow_sampler.clone()),
-                shadow_map1: (shadow1.clone(), shadow_sampler.clone()),
-                out_color: self.out_color.clone(),
-                out_depth: self.out_depth.clone(),
+            match *material {
```
alteous: This match expression is bloated. It may be worth splitting up the cases in the future.

src/factory.rs
```diff
@@ -181,20 +200,49 @@ impl Factory {
...
-        } else {
-            let f2i = |x: f32| I8Norm(cmp::min(cmp::max((x * 127.) as isize, -128), 127) as i8);
-            shape.vertices.iter().zip(shape.normals.iter()).map(|(v, n)| Vertex {
-                pos: [v.x, v.y, v.z, 1.0],
-                uv: [0.0, 0.0], //TODO
-                normal: [f2i(n.x), f2i(n.y), f2i(n.z), I8Norm(0)],
-            }).collect()
-        }
+        let position_iter = shape.vertices.iter();
+        let normal_iter: Box<Iterator<Item = [gfx::format::I8Norm; 4]>> = {
```
kvark: I'd prefer us to avoid `Box` here, especially wrapping the iterator, since that would mean a virtual dispatch per each iteration...

alteous: Would [`itertools::Either`](https://docs.rs/itertools/0.6.1/itertools/enum.Either.html) be a good alternative? We would have to rely on the compiler moving the loop condition out of the loop. There are a lot of branches to consider otherwise.

alteous: So the code would like this:

```rust
let position_iter = shape.vertices.iter();
let normal_iter = if shape.normals.is_empty() {
    Either::Left(iter::repeat(NORMAL_Z))
} else {
    Either::Right(shape.normals.iter().map(...))
};
// etc.

kvark: Sounds good

alteous:
Thanks - glad to be of service!

kvark:
Not too shabby for the first PR in this project! Please keep up :) 👍 👏

Issue #28, opened 2017-06-16
Shader code vendoring


kvark:
Current approach requires the user to copy the shaders folder, which is unfortunate. A proper alternative might be to include the shaders in the code (with `include_bin!`), plus allow overrides from the file system for fast iteration.

cc @alexander-irbis

alexander-irbis:
If you look back at three.js, the basic things there are already included in the package and the user begins to add something of his own, only when the basic capabilities begin to be missed. If you go further - three.js generates the shaders in accordance with the settings of renderer (but I don't know how useful this property is). On this side, the user who will not go into the mechanics of shaders, he just need to keep them up to date.

On the other hand, fast reload without restart a program is really sweet feature at the drafting stage. It will be very useful for advanced users. But again user need to remember about the need to merge changes from upstream if he go to change standard shaders.

Probably it would still be nice to separate standard shaders from custom ones, to enable and use standard shaders, and to develop your own.

kvark:
How about the following plan:
  - the renderer has all the shader code included from the source templates, an it potentially modifies it based on the user settings before linking
  - user may invoke a dump of the shaders to a specified folder (from the code, supposedly)
  - user may request a hot reload of the shaders from a specified folder. The engine may detect the change and re-load only changed shaders.

vitvakatu:
> user may invoke a dump of the shaders to a specified folder (from the code, supposedly)

How the user should use this opportunity? Create side application just to dump standard shaders (and change them then)? Or maybe dump and change them from code at runtime (sounds like disaster).

kvark:
@vitvakatu at runtime. It's not a disaster since this is not supposed to be used in production.

alexander-irbis:
I think the problem should be divided into two parts.

The developers of three-rs need to work with the shaders of the library. But they know where to get the shaders and what to do with them.

The user needs to develop his own shaders, for example, post-effects and may want to use the same mechanics to fast reload them.

These parts do not intersect directly with each other, but complement each other.

But there may be another scenario: user use third party libs, which will generate the necessary shaders and resources.

As a result, we get a loader with scoped updates from several places.

alexander-irbis:
And then, with sophisticated generators, possibility to dump current state of all generated stuff to explore and may be fix a resulting shaders' code will be very useful in itself.

alteous:
[...]

PR #67, opened 2017-08-02
Make it possible to run examples from anywhere in the source tree

nchashch:
Now you can call "cargo run --example <example name>" from any directory,
not only from the root "Cargo.toml" directory.

### DIFF
--- a/examples/anim.rs
+++ b/examples/anim.rs
@@ -93,7 +93,9 @@ fn make_geometry() -> three::Geometry {
 }
 
 fn main() {
-    let mut win = three::Window::new("Three-rs mesh blending example", "data/shaders").build();
+    let shaders_path: String = format!("{}/data/shaders", env!("CARGO_MANIFEST_DIR"));
+    let shaders_path_str: &str = shaders_path.as_str();
+    let mut win = three::Window::new("Three-rs mesh blending example", shaders_path_str).build();
     let mut cam = win.factory.perspective_camera(60.0, 1.0, 1000.0);
     cam.look_at([100.0, 0.0, 100.0], [0.0, 0.0, 30.0], Some([0.0, 1.0, 0.0].into()));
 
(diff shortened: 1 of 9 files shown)


alteous:
It might be a better idea to embed the shader code into the program instead using `include_bytes!` or otherwise.

alteous:
Unless the intention is that the shaders are overridable?

vitvakatu:
@alteous we need flexible solution, to let users change their shaders easily. See #28

kvark:
Thanks @nchashch !
We do need to change the way shaders are loaded - #28 
This PR solves a part of the issue, but sacrificing the initialization ergonomics. If we don't merge it, and focus on the proper solution, I don't think this is going to block anyone from playing with thee-rs, right?

nchashch:
TL;DR: I think hardcoding shaders and keeping this `format!(..., env!(...))` solution for assets would be ok.

Yes, this is a bit of a crutch. Note that apart from shaders '.obj' and '.gltf' assets in current examples have this "run from only root Cargo.toml directory" problem. It is a minor problem, so I don't think solving only this problem (without any other benefit to the user) will be worth any kind of significant engineering effort.

I still think that being able to run examples from anywhere in the source tree is important, because it will save frustration to people tinkering with them. This "unwrap error" because you are in a wrong directory can be quite confusing for a person new to Rust.

Are there any plans for asset management? Three-rs has ergonomics focus and is targeted at making simple things, so having brutalist bare bones asset management - loading files specified by hard coded filenames might make sense to keep the API simple.

kvark:
@nchashch 

> I think hardcoding shaders and keeping this format!(..., env!(...)) solution for assets would be ok.

I don't think this is what we want to end up with, but I'm fine with it being a temporary change.

> Are there any plans for asset management? 

Not at the moment, but we are open to the idea.

bors r+

PR #68, opened 2017-08-02
Relicense under dual MIT/Apache-2.0 License

### LINKED ISSUE #66: Relicense under dual MIT/Apache-2.0
TL;DR the Rust ecosystem is largely Apache-2.0. Being available under that
license is good for interoperation. The MIT license as an add-on can be nice
for GPLv2 projects to use your code.

# Why?
The MIT license requires reproducing countless copies of the same copyright
header with different names in the copyright field, for every MIT library in
use. The Apache license does not have this drawback. However, this is not the
primary motivation for me creating these issues. The Apache license also has
protections from patent trolls and an explicit contribution licensing clause.
However, the Apache license is incompatible with GPLv2. This is why Rust is
dual-licensed as MIT/Apache (the "primary" license being Apache, MIT only for
GPLv2 compat), and doing so would be wise for this project. This also makes
this crate suitable for inclusion and unrestricted sharing in the Rust
standard distribution and other projects using dual MIT/Apache.

Some ask, "Does this really apply to binary redistributions? Does MIT really
require reproducing the whole thing?" I'm not a lawyer, and I can't give legal
advice, but some Google Android apps include open source attributions using
this interpretation. Others also agree with it.
But, again, the copyright notice redistribution is not the primary motivation
for the dual-licensing. It's stronger protections to licensees and better
interoperation with the wider Rust ecosystem.

# Contributor checkoff
To agree to relicensing, comment with :
> I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.

Due to small count of contributors, let's do it without automatic scripts.
Leave comment here and I'll check your checkbox. 

# Checklist
- [x] @kvark 
- [x] @alteous 
- [x] @sunjay 
- [x] @bwasty 
- [x] @shmutalov  
- [x] @vitvakatu

vitvakatu:
Closes #66


kvark:
Thank you!
bors r+

PR #69, opened 2017-08-03
Custom material

kvark:


src/factory.rs
```diff
@@ -397,6 +400,35 @@ impl Factory {
...
+    pub fn basic_pipeline(&mut self,
+        shader_path: &str,
+        primitive: gfx::Primitive,
+        rasterizer: gfx::state::Rasterizer,
+        depth_state: gfx::state::Depth,
+    ) -> Result<BasicPipelineState, ()> {
+        let code_vs = get_shader(&self.root_shader_path, shader_path, "vs");
+        let code_ps = get_shader(&self.root_shader_path, shader_path, "ps");
+
+        let program = match self.backend.link_program(code_vs.as_bytes(), code_ps.as_bytes()) {
```
vitvakatu: Can be reduced with `unwrap_or_else(...)`, but it's not necessary.

src/factory.rs
```diff
@@ -397,6 +400,35 @@ impl Factory {
             target,
         }
     }
+
+    /// Create a basic mesh pipeline using a custom shader.
+    pub fn basic_pipeline(&mut self,
```
vitvakatu: I suppose this PR isn't finished yet? How can I use it in my program? We need an example.

src/render.rs
```diff
@@ -124,17 +127,20 @@ gfx_defines! {
     }
 }
 
-fn get_shader(root: &str, name: &str, variant: &str) -> String {
+pub fn get_shader(root: &str, raw_path: &str, variant: &str) -> String {
```
vitvakatu: `pub(crate)` maybe better here? (I think we need it in many places)

src/factory.rs
kvark: Yeah, an example would be nice. But I know this works because I use it in https://github.com/kvark/plane-splitter/blob/59573c6d414d1b9952e168b73e3f2f5a6c72711c/src/main.rs#L35
So we can have example usage in follow-ups.

src/factory.rs
kvark: This is not the case for `unwrap_or_else`, because the error branch actually returns an error.

src/render.rs
kvark: not strictly needed. `render` module is not public by itself, and this function is not re-exported, so it's not visible from the outside.

kvark:
bors r=vitvakatu

PR #70, opened 2017-08-04
Make default texel range to cover entire texture

vickenty:
Previous default was using color at the corner of the texture for the entire textured polygon, causing textured obj models to render incorrectly.


kvark:
Great addition, thank you!
bors r+

PR #71, opened 2017-08-09
Moare custom pipeline states

kvark:
Adds the blend & stencil states to custom pipelines.


kvark:
Ain't nobody caring
bors r+

Issue #65, opened 2017-08-01
Clean up math conversion code


kvark:
We had quite a few crappy manual conversions between `mint` and `cgmath` types. This is all obsolete now, fortunately, given that cgmath-0.15 has been published with native `mint` support.

shmutalov:
Oh, Zeus! Good news! Time to remove 
```rust
self.camera_q = key_q * self.camera_q;
self.camera_q = self.camera_q.normalize();

let rv: [f32; 3] = self.camera_q.v.into();
let rot = mint::Quaternion {
   s: self.camera_q.s,
   v: rv.into(),
};
```

crap!

PR #78, opened 2017-08-18
Update cgmath to version 0.15 with mint support

alteous:


vitvakatu:
Thank you for PR, @alteous 
But I think it's not necessary. I'm working on #65 now, updating `cgmath` will be included.
I don't mind, just fyi.

kvark:
agreed
bors r+

PR #81, opened 2017-08-23
Allow mesh instances to duplicate their materials

alteous:
There might be a more idiomatic way of providing this functionality. Would it be better to provide a `Mesh::material()` function instead?


vitvakatu:
We can provide `Mesh::geometry` and `Mesh::material` functions, I think it would be nice.
@kvark?

kvark:
We can't provide `geometry` since it's technically abstracted away for static meshes.
We can't provide `material` either, since it's only stored in the scenegraph, not mesh itself.

kvark:
thanks!
bors r+

PR #84, opened 2017-08-25
Add text module for ui text rendering.

vitvakatu:


src/lib.rs
```diff
@@ -190,6 +196,13 @@ impl Hub {
                         }
                     }
                 }
+                Operation::SetText(new_data) => {
+                    if let SubNode::UIText(ref mut data) = node.sub_node {
+                        data.section = new_data.section;
```
kvark: why not just `*mut data = new_data` ?

src/text.rs
```diff
@@ -0,0 +1,210 @@
+use std::path::PathBuf;
+use std::rc::Rc;
+use std::cell::RefCell;
+use std::fmt;
+
+use gfx_glyph;
+use gfx_glyph::{GlyphBrush, GlyphBrushBuilder, OwnedSection, Scale, StandardLineBreaker};
```
kvark: any reason for some of the types being imported but not the others in `gfx_glyph`?
I find it easier to just shorthand the alias, aka `use gfx_glyph as g`.

src/text.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    pub(crate) fn queue(&self, section: &OwnedSection, layout: Layout) {
+        let mut brush = self.brush.borrow_mut();
+        let layout: gfx_glyph::Layout<StandardLineBreaker> = layout.into();
+        brush.queue(section, &layout);
+    }
+
+    pub(crate) fn draw(&self,
+                       encoder: &mut Encoder<BackendResources, BackendCommandBuffer>,
```
kvark: let's use the newer style

src/text.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /// Change text layout.
+    /// Defaults to `Layout::SingleLine(Align::Left)`.
+    pub fn set_layout(&mut self, layout: Layout) {
+        self.data.layout = layout.clone();
+        let msg = Operation::SetText(self.data.clone());
+        let _ = self.object.tx.send((self.as_ref().downgrade(), msg));
+    }
+}
```
kvark: nit: newline

src/text.rs
```diff
@@ -0,0 +1,210 @@
...
+
+/// UI (on-screen) text.
+/// To use, create the new one using [`Factory::ui_text`](struct.Factory.html#method.ui_text)
+/// and add it to the scene using [`Scene::add`](struct.Scene.html#method.add).
+#[derive(Debug, Clone)]
+pub struct Text {
+    pub(crate) object: Object,
+    pub(crate) data: TextData,
```
kvark: note: carrying the copy of internal data with the local object is somewhat an anti-pattern, as you can see nothing else doing it, because otherwise animating those properties would be hard (w.r.t. synchronization between this local state and the remote state in the hub)

src/lib.rs
alteous: Also, `UIText` should be `UiText`.

src/lib.rs
```diff
@@ -85,6 +89,7 @@ struct LightData {
 #[derive(Debug)]
 enum SubNode {
     Empty,
+    UIText(TextData),
```
alteous: This should be called `UiText`.

vitvakatu: UI is abrreviation, but I don't mind.

src/text.rs
vitvakatu: Makes sense, yes.

src/lib.rs
vitvakatu: @kvark my mistake, will fix

src/text.rs
vitvakatu: Can you explain?

src/text.rs
vitvakatu: Ugh, I know. I can change it, but it will lead us to massive bunch of pattern matching in Hub.

src/text.rs
alteous: Like this:

```rust
pub(crate) fn draw(
    &self,
    encoder: &mut ...,
) -> ReturnType {
    ...
}

src/lib.rs
alteous: This is recommended by the [Rust API guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html) - see the note at the bottom of the table.

src/text.rs
vitvakatu: Nevermind, I will fix that.

vitvakatu:
Updated, don't forget to let me squash before merging.

vitvakatu:
Looks like I fixed everything, check it out, please.
Don't forget to let me squash before merging.

kvark:
homu delegate=alteous

kvark:
bors delegate=alteous

alteous:
Great PR - thank you!

bors r+

Issue #82, opened 2017-08-24
[RFC] Crate reorganization 


vitvakatu:
# Motivation
Three-rs is a very young project. The major part of it was developed in a very short period of time, that leads us to some structural problems.
The main problem - a lot of things is situated in the root module, while there are more suitable places. 

# New structure
- `camera` module 
    * struct `Camera` _moved from `lib.rs`_
    * trait `Projection`
    * struct `Orthographic`
    * struct `Perspective`
- `input` module - **add documentation**
- `window` module - no changes
- `scene` module
    * `Scene` struct
    * `Background` enum
- `render` struct 
_(?)_ rework `render` method to reduce nesting using dedicated private functions in match statements  
- `macros` module - no changes
- `factory` module
    * struct `Factory`
- `loader` module - **new module** _(?)_
    * struct `Loader`
    `Loader` is stored inside `Factory` struct and provides functions for loading textures, sounds, meshes and so on. This functionality is separated from `Factory`.
- `geometry` module - **new module**
    * struct `Geometry` _moved from `factory.rs`_
    * struct `GeometryShape` _moved from `factory.rs`_
- `mesh` module - **new module**
    * struct `Mesh` _moved from `lib.rs` and `scene.rs`_ 
    * struct `DynamicMesh` _moved from `lib.rs` and `scene.rs`_
    * common trait for both `Mesh` and `DynamicMesh` _(??)_
- `material` module - **new module** 
    * struct `Material` _moved from `scene.rs`_
    * struct `Texture`
- `lights` module - **new module**
    * structs `AmbientLight`, `PointLight`, `HemisphereLight`, `DirectionalLight` _moved from `scene.rs`_
- `node` module - **new module**
    * type `NodePointer`
    * type `Transform`
    * struct `Node`
    * struct `NodeInfo`
    * struct `NodeTransform`
- `object` module - **new module**
    * struct `Object`
    * struct `Group`
- `hub` module - **new module**
    * struct `Hub`
    * all types used by `Hub` and not noticed by me in other modules (`SubNode`, `LightData`, `GpuData`...)
- `sprite` module - **new module**
     * `Sprite` struct
- `controls` module - **new module** - not sure about the name
    * `orbit` submodule - contatins `OrbitalControls`
    * `animation` submodule - contains animation controls

#### Publicity
We should use `pub(crate)` to hide some things form user (not `#[doc(hidden)]`) and to allow access to fields and methods for other modules
#### Documentation
[...]

PR #86, opened 2017-08-26
Refactoring

vitvakatu:
See #82 for general information about changes.
Ready to merge


src/controls/orbit.rs
```diff
@@ -0,0 +1,107 @@
...
+use input::{Button, Input, MOUSE_LEFT};
+use node::Transform;
+use object::Object;
+
+/// Simple controls for Orbital Camera.
+///
+/// Camera is rotating around the fixed point without any restrictions.
+/// By default, it uses left mouse button as control button (hold it to rotate) and mouse wheel
+/// to adjust distance to the central point.
+pub struct OrbitControls {
```
alteous: The `Controls` suffix is redundant now.

src/geometry.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub tangents: Vec<mint::Vector4<f32>>,
+    /// Texture co-ordinates.
+    pub tex_coords: Vec<mint::Point2<f32>>,
+}
+
+impl Shape {
+    /// Create an empty shape.
+    pub fn empty() -> Self {
+        Shape {
+            vertices: Vec::new(),
```
alteous: FYI, `vec![]` does the same thing. It's faster to type. No need to change though.

vitvakatu: I know, but it's not my code actually. And I prefer `Vec::new()` as it more verbose.
There is no macro for `HashMap::new`, for example, so using `Vec::new` makes code a little bit more "solid"

src/input.rs
```diff
@@ -1,5 +1,3 @@
-#![allow(missing_docs)] //TODO
```
alteous: :+1:

src/controls/orbit.rs
vitvakatu: Makes sense, yes

src/input.rs
```diff
@@ -10,6 +8,7 @@ const PIXELS_PER_LINE: f32 = 38.0;
 
 pub type TimerDuration = f32;
 
+// TODO: Remove
 #[allow(dead_code)]
 struct InputState {
```
alteous: Can now be renamed `State`.

src/input.rs
```diff
@@ -27,10 +26,11 @@ struct InputDiff {
     mouse_wheel: Vec<f32>,
 }
 
+/// Controls user and system input from keyboard, mouse and system clock.
 pub struct Input(InputState, InputDiff);
```
alteous: `struct Input(State, Diff)`.

src/input.rs
```diff
@@ -59,32 +59,40 @@ impl Input {
...
 
+    /// Create new timer.
     pub fn time(&self) -> Timer {
         Timer {
             start: self.0.time_moment
         }
     }
 
+    /// Get current delta time (time since previous frame)
     pub fn get_time_delta(&self) -> f32 {
```
alteous: The `get_` prefixes should be removed.

src/input.rs
```diff
@@ -118,46 +126,56 @@ impl Input {
...
 pub const KEY_SPACE: Button = Button::Key(Key::Space);
+/// Left mouse button.
 pub const MOUSE_LEFT: Button = Button::Mouse(MouseButton::Left);
+/// Right mouse button.
 pub const MOUSE_RIGHT: Button = Button::Mouse(MouseButton::Right);
 
 impl Button {
+    /// Get the amount of hits (moves down) for this button.
+    /// You typically need to pass `window.input` as `input` parameter.
     pub fn count_hits(&self, input: &Input) -> u8 {
```
alteous: `hit_count`?

src/lib.rs
```diff
@@ -27,324 +27,44 @@ extern crate glutin;
...
 mod macros;
-mod camera;
+pub mod camera;
+pub mod controls;
+pub mod custom;
 mod factory;
+pub mod geometry;
+mod hub;
 mod input;
+pub mod lights;
```
alteous: `lights::Directional` vs `light::Directional`. I prefer the latter.

src/input.rs
vitvakatu: I prefer `delta_time`, actually.

src/input.rs
vitvakatu: Agreed

src/lib.rs
```diff
@@ -27,324 +27,44 @@ extern crate glutin;
 
 #[macro_use]
 mod macros;
-mod camera;
+pub mod camera;
+pub mod controls;
```
alteous: This plural is OK, because `controls::Orbit` makes more sense than `control::Orbit`.

src/lib.rs
vitvakatu: Should I rename or just reexport with different name?

src/lights.rs
```diff
@@ -0,0 +1,120 @@
...
+/// and [`DirectionalLight`](struct.DirectionalLight.html).
+#[derive(Clone, Debug)]
+pub struct ShadowMap {
+    pub(crate) resource: gfx::handle::ShaderResourceView<BackendResources, f32>,
+    pub(crate) target: gfx::handle::DepthStencilView<BackendResources, ShadowFormat>,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum ShadowProjection {
+    Ortho(Orthographic),
```
alteous: These names ought to agree. I prefer the more verbose `Orthographic`.

src/lights.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+impl ShadowMap {
+    #[doc(hidden)]
```
alteous: Is this intentional?

src/lights.rs
vitvakatu: Makes sense, if we'll have `Perspective` in future.

src/lights.rs
vitvakatu: Need to replace with `pub(crate)`, good catch

src/material.rs
```diff
@@ -0,0 +1,45 @@
...
+        wireframe: bool,
+    },
+    /// Lambertian diffuse reflection. This technique causes all closed polygons
+    /// (such as a triangle within a 3D mesh) to reflect light equally in all
+    /// directions when rendered.
+    MeshLambert { color: Color, flat: bool },
+    /// Material that uses Phong reflection model.
+    MeshPhong { color: Color, glossiness: f32 },
+    /// Physically-based rendering material.
+    MeshPbr {
```
alteous: three.js names their PBR material type `MeshStandardMaterial`. I used `Pbr` because it's shorter and more descriptive. Is this OK?

vitvakatu: I think it's okay. `MeshStandardMaterial` is too verbose (`Material::MeshStandardMaterial`)

src/object.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl Object {
+    /// Invisible objects are not rendered by cameras.
+    pub fn set_visible(&mut self, visible: bool) {
+        let msg = Operation::SetVisible(visible);
+        let _ = self.tx.send((self.node.downgrade(), msg));
+    }
+
+    /// Rotates object in the specific direction of `target`.
+    pub fn look_at<P>(&mut self, eye: P, target: P, up: Option<mint::Vector3<f32>>)
```
alteous: `eye` and `target` should have different type parameters. That can be fixed in another time though.

src/material.rs
vitvakatu: Moreover, we have `MeshBasic`. "Basic" and "Standard" - what to choose, huh?

src/object.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Set scale.
+    pub fn set_scale(&mut self, scale: f32) {
+        let msg = Operation::SetTransform(None, None, Some(scale));
+        let _ = self.tx.send((self.node.downgrade(), msg));
+    }
+
+    /// Get actual information about itself from the `scene`.
+    /// # Panics
+    /// Panics if `scene` doesn't have this `Object`.
+    pub fn sync(&mut self, scene: &Scene) -> NodeInfo {
```
alteous: What's the rationale behind the name `sync`? Is there any reason why it shouldn't be called `node_info` instead?

src/texture.rs
```diff
@@ -0,0 +1,60 @@
...
+    view: h::ShaderResourceView<BackendResources, T>,
+    sampler: h::Sampler<BackendResources>,
+    total_size: [u32; 2],
+    tex0: [f32; 2],
+    tex1: [f32; 2],
+}
+
+impl<T> Texture<T> {
+    #[doc(hidden)]
+    pub fn new(view: h::ShaderResourceView<BackendResources, T>,
```
alteous: Is this intentional?

src/texture.rs
```diff
@@ -0,0 +1,60 @@
...
+            self.total_size[1] as f32 - base.y as f32 - size.y as f32,
+        ];
+        self.tex1 = [
+            base.x as f32 + size.x as f32,
+            self.total_size[1] as f32 - base.y as f32,
+        ];
+    }
+
+    /// Returns normalized UV rectangle (x0, y0, x1, y1) of the current texel range.
+    pub fn get_uv_range(&self) -> [f32; 4] {
```
alteous: Should be `uv_range`.

src/object.rs
vitvakatu: `sync` is for synchronizing current Node state (by grabbing it from `Hub`). Not sure that user needs to know about it though

src/geometry.rs
alteous: :+1:

src/texture.rs
vitvakatu: Mistake, thanks

src/lib.rs
alteous: I meant the name of the module - `lights` vs `light`.

src/material.rs
alteous: I agree.

src/object.rs
alteous: If there's a performance penalty for using it due to synchronisation, then perhaps `sync` is the better name.

vitvakatu: Actually, yes, there is a performance penalty, but due to recalculation of scene state.

alteous:
Great work! Ready to merge I think.

src/controls/orbit.rs
```diff
@@ -0,0 +1,107 @@
...
+            let q_ver = Quaternion::from_angle_y(Rad(self.speed * (mouse_delta.x)));
+            let axis = self.transform.rot * Vector3::unit_x();
+            let q_hor = Quaternion::from_axis_angle(axis, Rad(self.speed * (mouse_delta.y)));
+            let post = Decomposed {
+                scale: 1.0 + input.mouse_wheel() / 1000.0,
+                rot: q_hor * q_ver,
+                disp: self.target.to_vec(),
+            };
+            self.transform = post.concat(&pre.concat(&self.transform));
+            let pf: mint::Vector3<f32> = self.transform.disp.into();
```
kvark: nit: easier to write with `from`: `let pf = mint::Vector3::from(self.transform.disp)`

src/controls/orbit.rs
```diff
@@ -0,0 +1,107 @@
...
+            self.object.set_transform(pf, self.transform.rot, 1.0);
+        }
+    }
+}
```
kvark: nit: newline missing, also in other files
I recommend configuring the editor to do this for ya

src/geometry.rs
```diff
@@ -0,0 +1,137 @@
...
+                .. Shape::empty()
+            },
+            .. Geometry::empty()
+        }
+    }
+
+    fn generate<P, G, Fpos, Fnor>(gen: G, fpos: Fpos, fnor: Fnor) -> Self where
+        P: EmitTriangles<Vertex=usize>,
+        G: IndexedPolygon<P> + SharedVertex<GenVertex>,
+        Fpos: Fn(GenVertex) -> mint::Point3<f32>,
```
kvark: this is not for your PR, but later on this would probably be cleaner and faster with a single function returning both position and normal

src/geometry.rs
```diff
@@ -0,0 +1,137 @@
...
+                           [pos[0] * 0.5 * sx, pos[1] * 0.5 * sy, pos[2] * 0.5 * sz].into()
+                       },
+                       |v| v.normal.into()
+        )
+    }
+
+    /// Create new Cylinder or Cone with desired top and bottom radius, height
+    /// and number of segments.
+    pub fn cylinder(radius_top: f32, radius_bottom: f32, height: f32,
+                        radius_segments: usize) -> Self
```
kvark: weird indentation

src/lib.rs
```diff
@@ -27,324 +27,44 @@ extern crate glutin;
...
-                AXIS_LEFT_RIGHT, AXIS_DOWN_UP};
+pub use controls::{KEY_SPACE, KEY_ESCAPE, MOUSE_LEFT, MOUSE_RIGHT, AXIS_DOWN_UP, AXIS_LEFT_RIGHT};
+pub use controls::{Button, Input, Timer, KeyAxis};
+pub use factory::Factory;
+pub use geometry::Geometry;
 pub use render::{ColorFormat, DepthFormat, Renderer, ShadowType, DebugQuadHandle};
-pub use scene::{Color, Background, Material, NodeTransform, NodeInfo,
-                Group, Sprite,
-                AmbientLight, DirectionalLight, HemisphereLight, PointLight};
+pub use scene::{Scene, Color, Background};
```
kvark: nit: re-order use statements

src/mesh.rs
```diff
@@ -0,0 +1,92 @@
+use geometry::Geometry;
+use hub::Operation;
+use material::Material;
+use object::Object;
+use render::DynamicData;
+
+
+/// [`Geometry`](struct.Geometry.html) with some [`Material`](struct.Material.html).
```
kvark: just wondering, is there a multi-line doc comment format?

alteous:
> More divergence from Three.js naming and signatures is unfortunate but inevitable.

This is probably best discussed elsewhere but do you see this as a problem? I think using Three.js as inspiration is a good idea since its design obviously 'works', however trying to mirror the JS (!) API exactly prevents the original design from being improved.

src/mesh.rs
vitvakatu: No, as far as I know

alteous: What do you mean by multi-line?

alteous: By the way, you can put the links at the end if preferable. I think this is recommended practice.

```rust
/// [`Geometry`] with some [`Material`].
///
/// [`Geometry`]: struct.Geometry.html
/// [`Material`]: struct.Material.html
struct Mesh;
```

PR #89, opened 2017-08-27
Multiple fixes.

### LINKED ISSUE #57: [Question] Builders API
At the moment `WindowBuilder` is constructed via `Window::new` method, that isn't common behavior for `new` methods. (see [clippy lint](https://github.com/Manishearth/rust-clippy/wiki#new_ret_no_self))
cc @kvark @shmutalov

vitvakatu:
Changelog:
1. Replace `Window::new` with `Window::builder` Closes #57 
2. Fix some of the remaining formatting issues after #86 
3. Now `mouse_delta`, `mouse_movements` and `mouse_pos` methods return coordinates in pixels. New methods `mouse_delta_ndc`, `mouse_movements_ndc` and `mouse_pos_ndc` added to retrieve NDC.
4. Rewriting of loading shaders in `factory` and `render` modules. Now functions use `Path` and `PathBuf` instead of `&str` and `String`. Also update `WindowBuilder` to accept generic parameters.
5. Rewrite `Renderer::map_to_ndc` to accept `mint::Point2<f32>` and make API more solid.
6. Fix after #86: export `WindowBuilder` and add documentation.
7. Add new `Window::size` method to retrieve window size in pixels.
UPD:
8. Rename `WindowBuilder` to `Builder`, export `window` module.


examples/pbr.rs
```diff
@@ -100,7 +100,7 @@ fn load_mesh(mesh: gltf::mesh::Mesh, factory: &mut three::Factory) -> three::Mes
             occlusion_strength: mat.occlusion_texture().map_or(1.0, |t| t.strength()),
             emissive_factor: mat.emissive_factor(),
             normal_scale: mat.normal_texture().map_or(1.0, |t| t.scale()),
-            
+
```
alteous: :innocent:

src/controls/orbit.rs
```diff
@@ -86,7 +86,7 @@ impl Orbit {
         }
 
         if input.mouse_movements().len() > 0 {
-            let mouse_delta = input.mouse_delta();
+            let mouse_delta = input.mouse_delta_ndc();
```
alteous: :+1:

examples/pbr.rs
vitvakatu: I just set up IDE as @kvark noted :wink:

src/factory.rs
```diff
@@ -84,14 +84,14 @@ fn f2i(x: f32) -> I8Norm {
...
-    pub fn new(mut backend: BackendFactory, shader_path: &str) -> Self {
+    pub fn new(mut backend: BackendFactory, shader_path: &Path) -> Self {
         let quad_buf = backend.create_vertex_buffer(&QUAD);
         let default_sampler = backend.create_sampler_linear();
         Factory {
             backend: backend,
             scene_id: 0,
             hub: Hub::new(),
-            root_shader_path: shader_path.to_string(),
+            root_shader_path: shader_path.to_owned(),
```
alteous: If you're going to call `to_owned` then it's probably best to take `PathBuf` instead of `&Path`.

alteous: Or better still, `Into<PathBuf>`.

src/window.rs
```diff
@@ -25,36 +28,43 @@ pub struct Window {
     pub scene: Scene,
 }
 
+/// Builder for creating new [`Window`](struct.Window.html) with desired parameters.
 pub struct WindowBuilder {
```
alteous: Can be called `Builder`.

src/window.rs
```diff
@@ -145,4 +156,10 @@ impl Window {
     pub fn render<P: Projection>(&mut self, camera: &Camera<P>) {
         self.renderer.render(&self.scene, camera);
     }
+
+    /// Get current window size in pixels.
+    pub fn size(&self) -> mint::Vector2<f32> {
```
alteous: :+1:

src/factory.rs
vitvakatu: Actually, it doesn't matter. `WindowBuilder::build` can't transfer own `PathBuf` so it needs either to pass `Path` or clone to the `Renderer::new`. `Renderer::new` uses this `Path` a lot, so `Path` is best choice to avoid cloning. After that, you need to pass it to `Factory::new` - so you need to call `to_owned` now or later - in `Factory::new`, as I did.

vitvakatu:
@kvark we lost both bors and Travis :(

alteous:
RIP.

src/input.rs
```diff
@@ -16,12 +16,14 @@ struct State {
     keys_pressed: HashSet<Key>,
     mouse_pressed: HashSet<MouseButton>,
     mouse_pos: mint::Point2<f32>,
+    mouse_pos_ndc: mint::Point2<f32>,
```
kvark: not sure why we'd want to have NDC and non-NC both tracked and exposed? The users can convert, if needed

vitvakatu: Converting needs access to window size. There are several ways to do so: 
1. Store window size inside input and convert to ndc on demand. 
2. Pass Renderer to Orbit::update
3. Move map_to_ndc function outside Renderer and pass window size to it as parameter

alteous: It's a small and useful feature that's hardly going to be a performance bottleneck. Why burden the user with a manual conversion?

alteous: It's needed for ray casting, for example.

vitvakatu: Agreed with alteous. It simplifies the API by holding all related stuff in one place.

vitvakatu:
Achtung, just found a bug in `aviator` example (related to the new input system). Forgot to update.

kvark:
Ok, I don't feel strongly about the NDC side of the input API. Let's proceed, since you guys do seem convinced.
bors r+

vitvakatu:
@kvark I understand your concern, but what we can do? Sometimes user doesn't have access to renderer, but it needs input in ndc.

PR #91, opened 2017-08-28
Add glTF support

alteous:
This PR implements loading of single glTF scenes. It is by no means a complete solution.

Test with `cargo run --release --example gltf` for the lantern model or `cargo run --release --example gltf -- test_data/SciFiHelmet.gltf` for the helmet model from the old PBR example.

#### Notable outstanding issues

- Cameras are ignored.
- Normals and tangents are not generated, ~~and the loader will panic if unavailable.~~
- The default scene is ignored because [I forgot to add the required function](https://github.com/alteous/gltf/issues/84).
- No animation support.
- Embedded image data is not decoded.

#### Non-issues

- If no scenes are available then nothing will be rendered. This is the correct behaviour according to the specification.

| Example model          | Does not panic | Renders OK | Animates OK | Notes
|------------------------|----------------|-------------------|--------------------|--|
| 2CylinderEngine        | ☒              | ☐                 | ☐                  | Not enough detail.
| AnimatedCube           | ☒              | ☒                 | ☐                  |
| AnimatedMorphCube      | ☒              | ☒                 | ☐                  |
| AnimatedMorphSphere    | ☒              | ☒                 | ☐                  |
| AnimatedTriangle       | ☒              | ☒                 | ☐                  |
| Avocado                | ☒              | ☒                 | ☒                  |
| BarramundiFish         | ☒              | ☒                 | ☒                  |
| BoomBox                | ☒              | ☒                 | ☒                  |
| Box                    | ☒              | ☒                 | ☒                  |
| BoxAnimated            | ☒              | ☒                 | ☐                  |
| BoxInterleaved         | ☒              | ☒                 | ☒                  |
| BoxTextured            | ☒              | ☒                 | ☒                  |
| BrainStem              | ☒              | ☒                 | ☐                  |
| Buggy                  | ☒              | ☒                 | ☐                  |
| Cameras                | ☐              | ☐                 | ☐                  |
| CesiumMan              | ☒              | ☒                 | ☒                  |
| CesiumMilkTruck        | ☒              | ☒                 | ☒                  |
| Corset                 | ☒              | ☒                 | ☒                  |
| Cube                   | ☒              | ☒                 | ☒                  |
| DamagedHelmet          | ☒              | ☒                 | ☒                  |
| Duck                   | ☒              | ☒                 | ☒                  |
| GearboxAssy            | ☒              | ☐                 | ☐                  | ???
| Lantern                | ☒              | ☒                 | ☒                  |
| MetalRoughSpheres      | ☒              | ☒                 | ☒                  | 
| Monster                | ☒              | ☐                 | ☐                  |
| NormalTangentTest      | ☒              | ☐                 | ☐                  | Needs normal generation, tangent generation, and back face culling switched off.
| ReciprocatingSaw       | ☒              | ☐                 | ☐                  |
| RiggedFigure           | ☒              | ☒                 | ☐                  |
| RiggedSimple           | ☒              | ☐                 | ☐                  |
| SciFiHelmet            | ☒              | ☒                 | ☐                  |
| SimpleMeshes           | ☒              | ☒                 | ☒                  |
| SmilingFace            | ☒              | ☒                 | ☒                  |
| Suzanne                | ☒              | ☒                 | ☒                  |
| TextureCoordinateTest  | ☒              | ☒                 | ☒                  | Facing away from camera.
| TextureSettingsTest    | ☒              | ☐                 | ☒                  | Needs back face culling switched off.
| Triangle               | ☐              | ☐                 | ☐                  | Ironically, the least complex model panics.
| TriangleWithoutIndices | ☒              | ☒                 | ☒                  |
| TwoSidedPlane          | ☒              | ☐                 | ☒                  | Needs back face culling switched off.
| VC                     | ☒              | ☐                 | ☐                  |
| WalkingLady            | ☒              | ☒                 | ☐                  |
| WaterBottle            | ☒              | ☒                 | ☒                  |


vitvakatu:
I'll be able to review in a few hours

alteous:
Ping @bwasty

src/factory.rs
```diff
@@ -543,6 +546,281 @@ impl Factory {
         self.request_texture(Path::new(path_str))
     }
 
+    /// Loads a `glTF` texture.
+    fn load_gltf_texture(
```
kvark: would be great to move all the `load_gltf` things out in a sub-module
you can still do `impl super::Factory {...}` from there

src/factory.rs
```diff
@@ -543,6 +546,281 @@ impl Factory {
...
+    /// Loads a single `glTF` node.
+    fn load_gltf_node(
+        &mut self,
+        node: &gltf::Node,
+        buffers: &gltf_importer::Buffers,
+        base: &Path,
+        meshes: &mut VecMap<Vec<Mesh>>,
+        instances: &mut Vec<Mesh>,
+    ) -> Group {
+        fn recurse(
```
kvark: might look cleaner if you unwrap the recursion by using a stack container explicitly

vitvakatu: I would like to avoid recursion if it's possible. Rust doesn't have TCO, so it can be the source of problems.

src/factory.rs
```diff
@@ -543,6 +546,281 @@ impl Factory {
...
+            node: &gltf::Node,
+            buffers: &gltf_importer::Buffers,
+            base: &Path,
+            parent: &mut Group,
+            meshes: &mut VecMap<Vec<Mesh>>,
+            instances: &mut Vec<Mesh>,
+            factory: &mut Factory,
+        ) {
+            let mut group = factory.group();
+            // Groups do not handle non-uniform scaling, so for now
```
vitvakatu: Add `TODO:` label, please. It would help us to find this place in future.

kvark:
I think 10Mb is nothing, assuming we aren't going to move them around too much ;)
If we need more, we can look into https://git-lfs.github.com/

vitvakatu:
@kvark yes, probably. It wasn't demand, just suggestion.

src/factory.rs
alteous: I'll redesign.

alteous:
@kvark: I adapted the code from your group example. The code is more elegant now. Thanks for the tip.

alteous:
Normal and tangent generation is coming next.

alteous:
There are significant changes to the camera API. Let me know what you think.

src/camera.rs
```diff
@@ -1,43 +1,89 @@
...
+
     /// Projection parameters of this camera.
-    pub projection: P,
+    pub projection: Projection,
 }
 
-impl<P> AsRef<NodePointer> for Camera<P> {
+impl Camera {
+    /// Computes the projection matrix representing the camera's projection.
+    pub fn project(&self, aspect_ratio: f32) -> mint::ColumnMatrix4<f32> {
```
vitvakatu: Actually, `get_matrix` is more obvious for me. Any reason for that naming?

examples/gltf.rs
```diff
@@ -80,7 +83,7 @@ fn main() {
         }
         let yrot = cgmath::Quaternion::<f32>::from_angle_y(cgmath::Rad(-st.yaw));
         let xrot = cgmath::Quaternion::<f32>::from_angle_x(cgmath::Rad(-st.pitch));
-        cam.set_transform(cgmath::Point3::from_vec(st.position), yrot * xrot, 1.0);
```
alteous: Time to write first person controls.

src/camera.rs
```diff
@@ -1,43 +1,89 @@
...
+            center: center.into(),
+            extent_y,
+            near,
+            far,
+        })
+    }
+
+    /// Constructs a perspective projection.
+    pub fn perspective(
+        fov_y: f32,
```
alteous: I ran into a degrees vs. radians problem here. It might be worth altering the API to avoid in the future.

src/factory/load_gltf.rs
```diff
@@ -275,23 +304,29 @@ impl super::Factory {
...
         let default = Path::new("");
         let base = path.parent().unwrap_or(&default);
         let (gltf, buffers) = gltf_importer::import(path).expect("invalid glTF 2.0");
+        let mut cameras = Vec::new();
         let mut meshes = VecMap::new();
         let mut instances = Vec::new();
         let mut group = self.group();
 
-        if let Some(scene) = gltf.scenes().nth(0) {
+        if let Some(scene) = gltf.default_scene() {
```
alteous: Will be available in gltf 0.9.1.

src/camera.rs
vitvakatu: Agreed, smth like `type Deg=f32` and `type Rad=f32` may be enough.

src/factory/load_gltf.rs
vitvakatu: Great!

examples/gltf.rs
vitvakatu: There is one problem... I'll explain in gitter

src/camera.rs
alteous: I renamed once and forgot to rename again. `Camera::matrix` is probably a better name, would you agree?

alteous: Or even, `Camera::compute_projection_matrix`.

vitvakatu: I prefer `matrix` or `projection_matrix`

src/camera.rs
```diff
@@ -1,43 +1,89 @@
...
+        near: f32,
```
kvark: nit: we could use `Range<f32>`

alteous: Actually, I've missed something here. glTF allows for an infinite projection with a near value but no far value. How should we expose this?

Possibilities
1. Use `Option`
   * `fn perspective(fov_y: f32, near: f32, far: Option<f32>)`
   * `fn orthographic(..., range: Range<f32>)` - inconsistent
2. Separate functions for each case
    * `fn finite_perspective(fov_y: f32, range: Range<f32>)`
    * `fn infinite_perspective(fov_y: f32, range: RangeFrom<f32>)`
    * `fn orthographic(..., range: Range<f32>)`
3. Make generic
    * `fn perspective<R: Into<PerspectiveRange>>(fov_y: f32, range: R)`
    * `fn orthographic(..., range: Range<f32>)`

alteous: I went with (3). Let me know what you think. :smile:

alteous:
I'd like to start working on animations but since there will be a lot of changes it might be best to start another PR.

vitvakatu:
Yes, Animation should be in separate PR

alteous:
Will squashing be necessary before merging?

kvark:
To a reasonable degree. Mainly, to have each  commit to build for bisection, but not too small.

Cargo.toml
```diff
@@ -24,14 +24,15 @@ froggy = "0.4"
 genmesh = "0.5"
 gfx = "0.16"
 gfx_glyph = "0.3"
-gltf = "0.9"
-gltf-importer = "0.9"
-gltf-utils = "0.1"
+gltf = { path = "../gltf", version = "0.9.1" }
```
vitvakatu: Don't forget to change it

alteous:
Should be ready to merge now as 'partial support.

kvark:
Sweet!
bors r+

PR #97, opened 2017-09-08
First person camera improvements & raw mouse input

vitvakatu:


examples/obj.rs
```diff
@@ -11,11 +11,10 @@ fn main() {
...
     let cam = win.factory.perspective_camera(60.0, 1.0 .. 10.0);
-    let mut controls = three::controls::Orbit::builder(
-        &cam,
-        [0.0, 2.0, -5.0],
-        [0.0, 0.0, 0.0]
-    ).build();
+    let mut controls = three::controls::Orbit::builder(&cam)
+        .position([0.0, 2.0, -5.0])
+        .target([0.0, 0.0, 0.0])
+        .build();
```
alteous: This is much better.

src/controls/first_person.rs
```diff
@@ -1,11 +1,33 @@
 use cgmath;
 use mint;
-use std;
 
 use cgmath::Rotation3;
 use input::{Button, Input, Key};
 use object::Object;
 
+#[derive(Clone)]
+struct FirstPersonKeys {
```
alteous: This could just be called `Keys` since it's already prefixed by the module name (`first_person::Keys` vs `first_person::FirstPersonKeys`).

src/input.rs
```diff
@@ -47,6 +48,7 @@ impl Input {
             keys_hit: Vec::new(),
             mouse_moves: Vec::new(),
             mouse_moves_ndc: Vec::new(),
+            mouse_moves_raw: Vec::new(),
```
alteous: What's the difference between `mouse_moves` and `mouse_moves_raw`?

src/controls/first_person.rs
vitvakatu: Uh oh, thanks, I always forget about it

src/input.rs
vitvakatu: Raw input isn't restricted by window and it doesn't relate on screen pixels/window size.
Moreover, it's not always mouse input, it can be any device. I tried to notice that facts in docs.

alteous:
Great PR! I'm looking forward to trying this out. 😄

src/controls/first_person.rs
```diff
@@ -1,11 +1,33 @@
...
 use mint;
-use std;
 
 use cgmath::Rotation3;
 use input::{Button, Input, Key};
 use object::Object;
 
+#[derive(Clone)]
+struct Keys {
+    pub forward: Option<Button>,
```
kvark: any reason this can't use `KeyAxis`? it's not like you are going to map the forward/backward (for example) to mouse buttons anyway

src/window.rs
```diff
@@ -146,6 +146,19 @@ impl Window {
                     MouseWheel { delta, .. } => input.mouse_wheel_input(delta),
                     _ => {}
                 },
+                glutin::Event::DeviceEvent { event, .. } => match event {
+                    glutin::DeviceEvent::Motion { axis, value } => {
+                        let delta: mint::Vector2<f32> = if axis == 0 {
```
kvark: isn't `delta` type inferred here?

src/controls/first_person.rs
vitvakatu: good point, I'll change this to `KeyAxis`

src/window.rs
vitvakatu: I'll check it

src/controls/first_person.rs
vitvakatu: What naming should I use?

src/window.rs
```diff
@@ -146,6 +146,19 @@ impl Window {
                     MouseWheel { delta, .. } => input.mouse_wheel_input(delta),
                     _ => {}
                 },
+                glutin::Event::DeviceEvent { event, .. } => match event {
+                    glutin::DeviceEvent::Motion { axis, value } => {
+                        let delta = if axis == 0 {
```
kvark: nit: could be nicer with just `let delta = match axis {...};`

PR #98, opened 2017-09-08
add show_cursor function (set visibility cursor)

wallstbrok:


shmutalov:
Cool

alteous:
Thanks for the PR! Looks good to me.

kvark:
bors r+
btw, @alteous and @vitvakatu - you don't need to wait for me if there is a consensus between both of you and not many concerns. Just signal bors to merge things like this next time.

PR #99, opened 2017-09-10
Basic audio support

vitvakatu:
Also fix warnings in window.rs and remove unused `winit` dependency.


vitvakatu:
Need to add `alsa-sys` lib to travis...

src/lib.rs
```diff
@@ -51,6 +52,7 @@ mod texture;
 #[cfg(feature = "opengl")]
 pub mod window;
 
+pub use audio::{Clip, Source};
```
alteous: It might not be a good idea to export these since it's not clear what `Clip` and `Source` mean on their own. In the upcoming code for animations, there is also a `Clip` type which will clash.

vitvakatu: Reasonable

vitvakatu:
Strange, travis fails on stable...

vitvakatu:
Checked locally on stable, no problems

vitvakatu:
Oh, it fails on nightly, not on stable...

vitvakatu:
Okay, if there are any concerns we can always fix them later.
Let's try to merge it. (hah, I just want to play with bors)

bors r+

PR #100, opened 2017-09-11
Add textured background

### LINKED ISSUE #101: Textured scene background

vitvakatu:
Closes #101


vitvakatu:
bors r=alteous

kvark:
Please keep in mind that we are wasting fillrate here. The background texture should be drawn between the opaque and the transparent passes, using depth test to avoid touching pixels already drawn.

vitvakatu:
@kvark got it, will fix soon

PR #103, opened 2017-09-11
Animations part 1 of 2

alteous:
This PR implements the first part of the animation system targeting node transforms.

#### Implemented features

* Set of `Action`, `Mixer`, and `Clip` structs, inspired by the three.js animation system.
* Keyframe targets for simple binding pairs.
  * `(Position, Vector3)`
  * `(Orientation, Quaternion)`,
  * `(Orientation, Euler)`
  * `(Scale, Scalar)`
* Linear interpolation between keyframes.
* Support for node animations in the glTF loader.
* Two loop modes, namely `Once` and `Repeat`.

#### Planned features for part 2/2

* Mesh skinning and morph targets.
* Keyframe targets for the above.
* Implementation of `Discrete`, `Cubic`, and `CatmullRom` interpolation modes.
* Implementation of `PingPong` loop mode.
* Complete support for glTF animations.


alteous:
I've made a PR so everyone can track the progress and add their input along the way.

examples/gltf-animation.rs
```diff
@@ -0,0 +1,39 @@
...
+    let shaders = concat!(env!("CARGO_MANIFEST_DIR"), "/data/shaders");
+    let mut win = three::Window::builder("Three-rs glTF example", &shaders).build();
+    let mut light = win.factory.directional_light(0xFFFFFF, 7.0);
+    light.look_at([1.0, 1.0, 1.0], [0.0, 0.0, 0.0], None);
+    win.scene.add(&light);
+    win.scene.background = three::Background::Color(0xC6F0FF);
+
+    let default = concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/Lantern.gltf");
+    let path = std::env::args().nth(1).unwrap_or(default.into());
+    let (group, mut cameras, _meshes, clips) = win.factory.load_gltf(&path);
```
vitvakatu: I'm voting for new struct. Tuples are cool, but not in such situations.

run
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+CARGO_INCREMENTAL=1 cargo run --example gltf-animation -- ~/gltf/glTF-Sample-Models/2.0/$1/glTF/$1.gltf
```
vitvakatu: What's the purpose of this file? I hope it's just helper for you and won't be exposed to the user.

alteous: It's just for me. I swear I added it to .gitignore...

examples/gltf-animation.rs
alteous: I'm inclined to agree. What does @kvark think?

src/animation.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    #[derive(Clone, Debug)]
+    pub enum Target {
+        Orientation(Frames<mint::Quaternion<f32>>),
+        Position(Frames<mint::Point3<f32>>),
+        Scale(Frames<f32>),
+    }
+
+    #[derive(Clone, Debug)]
+    pub struct Track {
```
alteous: This corresponds to three.js's [`KeyframeTrack`](https://threejs.org/docs/#api/animation/KeyframeTrack) type.

src/animation.rs
```diff
@@ -0,0 +1,151 @@
...
+
+            let frame_start = self.values[i];
+            let frame_end = self.values[i + 1];
+            let s = t - self.times[i];
+            frame_start * (1.0 - s) + frame_end * s
+        }
+    }
+
+    #[derive(Clone, Debug)]
+    pub enum Target {
```
alteous: three.js uses `PropertyBinding` and is a bit more flexible in terms of what the animation can target. I've made this enum for simplicity.

src/animation.rs
```diff
@@ -0,0 +1,151 @@
...
+
+pub mod keyframe {
+    use cgmath;
+    use mint;
+    use super::Interpolation;
+    use Object;
+
+    /// The frames of a target.
+    #[derive(Clone, Debug)]
+    pub struct Frames<T> {
```
alteous: I'm trying to mimic the [`times` / `values`](https://threejs.org/docs/#api/animation/KeyframeTrack.times) properties in a type-safe way here.

src/animation.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub enum Ending {
+    ZeroCurvature,
+    ZeroSlope,
+    WrapAround,
+}
+
+#[derive(Clone, Debug)]
+pub struct Loop {
```
alteous: This is temporary, just for testing at the moment.

alteous:
I'm quite confused to how [`AnimationMixer`](https://threejs.org/docs/#api/animation/AnimationMixer) and [`AnimationAction`](https://threejs.org/docs/#api/animation/AnimationAction) work. They seem to depend on each other. :worried:

src/animation.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[derive(Clone, Debug)]
+    pub enum Target {
+        Orientation(Frames<mint::Quaternion<f32>>),
+        Position(Frames<mint::Point3<f32>>),
+        Scale(Frames<f32>),
+    }
+
+    #[derive(Clone, Debug)]
+    pub struct Track {
+        pub object: Object,
```
alteous: Although it works, I'm not sure if this is the correct implementation. I'm confused to why `PropertyBinding` exists but one has to specify the root object in an `AnimationMixer`.

alteous:
Check out `cargo run --example gltf-animation`! :smile:

alteous:
Crazy idea: We could (ab)use futures to schedule animations.

```rust
let animation = get_up()
    .and_then(|| walk_around())
    .and_then(|| sit_down());
animation.play(); // or animation.update()
```

kvark:
@alteous I'd much welcome us relying on futures here instead of building our own animation graphs. Looking forward to see if you get any luck with it 👍

alteous:
The use of futures to build animation graphs can be saved for another PR.

alteous:
All that's left is to write some decent documentation and cleanup.

src/factory/load_gltf.rs
```diff
@@ -269,22 +273,98 @@ impl super::Factory {
...
+                    None => continue,
+                };
+                let input = sampler.input();
+                let output = sampler.output();
+                use gltf::animation::InterpolationAlgorithm::*;
+                let interpolation = match sampler.interpolation() {
+                    Linear => animation::Interpolation::Linear,
+                    Step => animation::Interpolation::Discrete,
+                    CubicSpline => animation::Interpolation::Smooth,
+                    CatmullRomSpline => unimplemented!(),
```
alteous: Should be an easy fix.

src/factory/load_gltf.rs
```diff
@@ -319,6 +403,6 @@ impl super::Factory {
             meshes.insert(i, instances);
         }
 
-        (group, cameras, meshes)
+        (group, cameras, meshes, clips)
```
alteous: Shall we return a `Gltf` struct instead?

```rust
struct Gltf {
    cameras: Vec<Camera>,
    clips: Vec<Clip>,
    meshes: Vec<Mesh>,
    scenes: Vec<Group>,
}
```

src/factory/load_gltf.rs
```diff
@@ -293,6 +373,8 @@ impl super::Factory {
         let mut cameras = Vec::new();
         let mut meshes = VecMap::new();
         let mut instances = Vec::new();
+        let mut node_map = HashMap::new();
+        let mut clips = Vec::new();
         let mut group = self.group();
 
         if let Some(scene) = gltf.default_scene() {
```
alteous: Was thinking `gltf.default_scene().or(gltf.scenes.nth(0))`. Opinions?

vitvakatu: Agreed

src/factory/load_gltf.rs
vitvakatu: I think it's better, but not the necessity

alteous:
Ready to go!

alteous:
Ready to merge. :smile:

src/animation.rs
```diff
@@ -0,0 +1,503 @@
...
+    rx: mpsc::Receiver<Message>,
+    tx: mpsc::Sender<Message>,
+}
+
+impl Action {
+    fn send(
+        &mut self,
+        operation: Operation,
+    ) -> &mut Self {
+        let message = (self.pointer.downgrade(), operation);
```
alteous: Did I implement this correctly? The receiving end is at `Mixer::update`.

Cargo.toml
```diff
@@ -24,9 +24,9 @@ froggy = "0.4"
 genmesh = "0.5"
 gfx = "0.16"
 gfx_glyph = "0.3"
-gltf = "0.9.1"
-gltf-importer = "0.9.1"
-gltf-utils = "0.9.1"
+gltf = { features = ["names"], version = "0.9.2" }
+gltf-importer = { features = ["names"], version = "0.9.2" }
```
kvark: I wonder if `gltf` can follow `serde` an expose external crates in public via features (e.g `features = ["names", "importer", "utils"]`)

examples/gltf-animation.rs
```diff
@@ -0,0 +1,28 @@
+extern crate three;
+
+fn main() {
+    let title = "Three-rs glTF animation example";
+    let shaders = concat!(env!("CARGO_MANIFEST_DIR"), "/data/shaders");
```
kvark: we should really get the shader problem solved as the highest priority
if not someone else, I'll look into it this week

src/animation.rs
```diff
@@ -0,0 +1,503 @@
...
+//! The created actions are enabled by default in the 'play' state. This means that
+//! when calling [`Mixer::update`] the created actions will begin to be played back
+//! immediately.
+//!
+//! ```rust,no_run
+//! # let mut window = three::Window::builder("", "").build();
+//! # let mut mixer = three::animation::Mixer::new();
+//! # let (group, _, _, clips) = window.factory.load_gltf("AnimatedScene.gltf");
+//! # window.scene.add(&group);
+//! let actions: Vec<three::animation::Action> = clips
```
kvark: no need to specify the type here, I prefer doing `.collect::<Vec<_>>()` at the end instead

src/animation.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// Play the clip with the chosen number of repetitions, alternatively
+    /// playing in forward and reverse order.
+    PingPong(Repetitions),
+}
+
+/// Describes the number of times an [`Action`] repeats itself.
+///
+/// [`Action`]: struct.Action.html
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub enum Repetitions {
```
kvark: not 100% we need a new type for that, as opposed to having something like 
```rust
Repeat { repetitions: Option<usize> },
```

src/animation.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// Repeat an exact number of times.
+    Finite(i32),
```
kvark: why `i32`?

src/animation.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// [`Vector3`]: enum.Values.html#variant.Vector3
+    Position,
+
+    /// Targets the orientation property of an [`Object`].
+    ///
+    /// The corresponding keyframe values must be [`Quaternion`].
+    ///
+    /// [`Object`]: ../object/struct.Object.html
+    /// [`Quaternion`]: enum.Values.html#variant.Quaternion
+    Orientation,
```
kvark: should we differentiate between quaternion binding and euler binding?

src/animation.rs
```diff
@@ -0,0 +1,503 @@
...
+    ///
+    /// [`Object`]: ../object/struct.Object.html
+    /// [`Scalar`]: enum.Values.html#variant.Scalar
+    Scale,
+}
+
+/// An index into the frames of a track.
+enum Frame {
+    /// The time is before the start of the frames.
+    Unstarted,
```
kvark: the comments seem to refer to something different from `Frame`. I believe the `enum` needs to be renamed for clarity.

src/animation.rs
kvark: looks fine to me

src/animation.rs
```diff
@@ -0,0 +1,503 @@
...
+        &mut self,
+        delta_time: f32,
+    ) {
+        if self.paused || !self.enabled {
+            return;
+        }
+
+        self.local_time += delta_time * self.local_time_scale;
+        let mut finish_count = 0;
+        for mut track in &mut self.clip.tracks {
```
kvark: why do we mutate the clip tracks?

src/factory/load_gltf.rs
```diff
@@ -269,22 +276,106 @@ impl super::Factory {
...
+                let target = channel.target();
+                let node = target.node();
+                let object = match node_map.get(&node.index()) {
+                    Some(object) => object.clone(),
+                    // This animation does not correspond to any loaded node.
+                    None => continue,
+                };
+                let input = sampler.input();
+                let output = sampler.output();
+                use gltf::animation::InterpolationAlgorithm::*;
```
kvark: please move all use statements up to the beginning of the block/function

src/factory/load_gltf.rs
```diff
@@ -269,22 +276,106 @@ impl super::Factory {
...
+                    CubicSpline => animation::Interpolation::Cubic,
+                    CatmullRomSpline => animation::Interpolation::CatmullRom,
+                };
+                use animation::{Binding, Track, Values};
+                let times: Vec<f32> = AccessorIter::new(input, buffers).collect();
+                let (binding, values) = match target.path() {
+                    gltf::animation::TrsProperty::Translation => {
+                        let values = AccessorIter::<[f32; 3]>::new(output, buffers)
+                            .map(|v| {
+                                mint::Vector3 {
```
kvark: can we just use `mint::Vector3::from`?

src/factory/load_gltf.rs
```diff
@@ -269,22 +276,106 @@ impl super::Factory {
...
+                                }
+                            })
+                            .collect::<Vec<_>>();
+                        assert_eq!(values.len(), times.len());
+                        (Binding::Position, Values::Vector3(values))
+                    }
+                    gltf::animation::TrsProperty::Rotation => {
+                        let values = AccessorIter::<[f32; 4]>::new(output, buffers)
+                            .map(|r| {
+                                mint::Quaternion {
```
kvark: similarly, can we use `mint::Quaternion::from`?

src/factory/load_gltf.rs
```diff
@@ -269,22 +276,106 @@ impl super::Factory {
...
+                                        z: r[2],
+                                    },
+                                }
+                            })
+                            .collect::<Vec<_>>();
+                        assert_eq!(values.len(), times.len());
+                        (Binding::Orientation, Values::Quaternion(values))
+                    }
+                    gltf::animation::TrsProperty::Scale => {
+                        let values = AccessorIter::<[f32; 3]>::new(output, buffers)
```
kvark: needs a warning that non-uniform scales are mishandled

src/animation.rs
alteous: The intention is to be explicit as this is documentation.

examples/gltf-animation.rs
alteous: Totally agree, I'm working on this now. :smile:

Cargo.toml
alteous: That's a pretty good idea. I'll make an issue for that.

src/animation.rs
alteous: Why not an `i32`? :smile:

src/animation.rs
alteous: That's a good point. I'm tempted to not support Euler angles due to the possibility of gimbal lock. Do you think it's an important feature to have?

src/animation.rs
alteous: I'm really not sure what to name this. Coming up with the enum variant names was hard enough. It represents which keyframe a particular local time lies between. Name suggestions are most welcome.

src/animation.rs
alteous: Looks like an oversight. I'll change that.

src/factory/load_gltf.rs
alteous: Yes. Will fix.

src/factory/load_gltf.rs
alteous: I didn't want to risk it. :laughing:

src/factory/load_gltf.rs
alteous: Does that mean a `warn!` log or a note in the docs (or both)?

src/animation.rs
kvark: I'll agree as soon as you explain the semantics of negative repetition count to me ;)

src/animation.rs
kvark: Yes, some 3D suites / animator workflows would produce euler angle clips. I don't think gimbal lock is a concern here - the 3D suite can make sure that the produced clips are valid, it's not our task to validate.

src/animation.rs
kvark: how about `FrameRef`?

src/factory/load_gltf.rs
kvark: Totally get it, but we should encourage `mint` conversions and standards for the greater ~~god~~ good.

src/factory/load_gltf.rs
kvark: I'm afraid a `warn` repeated thousands of time would be troublesome. Just a comment would be fine for now.

src/animation.rs
alteous: It's the number of repetitions back in time, duh!

...I'll change it back to `u32`. :smile:

src/animation.rs
alteous: I prefer the other way as the intention is more clear. Also, this pattern is recommended by the [API guidelines](https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#arguments-convey-meaning-through-types-not-bool-or-option-c-custom-type).

kvark: I don't see that API guidelines section applicable here. We aren't talking about function arguments, where `bool` indeed becomes ambiguous for the lost semantics. In our case, there is always `repetitions` field involved, meaning the semantics is clear.

alteous: If we call the field `max_repetitions: Option<usize>` then it would make more sense.

src/animation.rs
alteous: It's not an oversight. It's because `Track` holds the `Object` it updates. I'm not sure how else to implement this.

src/animation.rs
alteous: I called it `limit` because that's shorter.

src/factory/load_gltf.rs
```diff
@@ -269,22 +276,93 @@ impl super::Factory {
...
+                    gltf::animation::TrsProperty::Scale => {
+                        // TODO: Groups do not handle non-uniform scaling, so for now
+                        // we'll choose Y to be the scale factor in all directions.
+                        let values = AccessorIter::<[f32; 3]>::new(output, buffers)
+                            .map(|s| s[1])
+                            .collect::<Vec<_>>();
+                        assert_eq!(values.len(), times.len());
+                        (Binding::Scale, Values::Scalar(values))
+                    }
+                    gltf::animation::TrsProperty::Weights => unimplemented!(),
```
alteous: This will be the target of our mesh animations in the next animation PR.

src/animation.rs
```diff
@@ -0,0 +1,532 @@
...
+pub enum LoopMode {
+    /// Play the clip in forward order exactly once, i.e. do not loop at all.
+    Once,
+
+    /// Play the clip in forward order, repeating from the start.
+    Repeat {
+        /// The maximum number of repetitions.
+        ///
+        /// When set to `None`, the loop will repeat indefinately.
+        limit: Option<u32>,
```
alteous: Removed `Repetitions` in favour of `limit: Option<u32>`.

src/animation.rs
```diff
@@ -0,0 +1,532 @@
...
+    ///
+    /// The corresponding keyframe values must be [`Vector3`].
+    ///
+    /// [`Object`]: ../object/struct.Object.html
+    /// [`Vector3`]: enum.Values.html#variant.Vector3
+    Position,
+
+    /// Targets the orientation property of an [`Object`].
+    ///
+    /// The corresponding keyframe values must be [`Quaternion`] or [`Euler`].
```
alteous: N.B.

src/animation.rs
```diff
@@ -0,0 +1,532 @@
...
+    Ended,
+}
+
+/// The keyframe values of a [`Track`].
+///
+/// [`Track`]: struct.Track.html
+#[derive(Clone, Debug)]
+pub enum Values {
+    /// Euler angle keyframes in radians.
+    Euler(Vec<mint::EulerAngles<f32, IntraXyz>>),
```
alteous: Corresponds to `cgmath::EulerAngles`. Is this the most common representation?

src/animation.rs
```diff
@@ -0,0 +1,532 @@
...
+            match (track.binding, &track.values) {
+                (Binding::Orientation, &Values::Euler(ref values)) => {
+                    let frame_start_value = {
+                        let euler = values[frame_index];
+                        cgmath::Quaternion::from(
+                            cgmath::Euler::new(
+                                cgmath::Rad(euler.a),
+                                cgmath::Rad(euler.b),
+                                cgmath::Rad(euler.c),
+                            ),
```
alteous: If there's a better way please let me know.

src/animation.rs
```diff
@@ -0,0 +1,532 @@
...
+                        )
+                    };
+                    let update = frame_start_value.slerp(frame_end_value, s);
+                    track.object.set_orientation(update);
```
alteous: We mutate `track.object` here, which is why `&mut self.clip.tracks` is required.

src/animation.rs
```diff
@@ -0,0 +1,532 @@
...
+                LoopMode::Once => self.enabled = false,
+                LoopMode::Repeat { limit: None } => self.local_time = 0.0,
+                LoopMode::Repeat { limit: Some(0) } => self.enabled = false,
+                LoopMode::Repeat { limit: Some(n) } => {
+                    self.local_time = 0.0;
+                    self.loop_mode = LoopMode::Repeat { limit: Some(n - 1) };
+                }
+                LoopMode::PingPong { .. } => {
+                    // TODO
+                    unimplemented!()
```
alteous: Leaving for the next animation PR.

src/animation.rs
kvark: Can't answer that right now. Let's revisit it later.

src/animation.rs
kvark: We need a way to separate immutable stuff from mutable...

src/animation.rs
```diff
@@ -0,0 +1,532 @@
...
+}
+
+/// A reusable set of keyframe tracks which represent an animation.
+#[derive(Clone, Debug)]
+pub struct Clip {
+    /// A name for this clip.
+    pub name: Option<String>,
+
+    /// The animation keyframe tracks.
+    pub tracks: Vec<Track>,
```
kvark: can we have `pub tracks: Vec<(Track, Object)>` so that object mutability is moved out of the track data?

alteous: Sure.

Issue #31, opened 2017-06-16
Material definition


kvark:
Current approach with materials being an `enum` is not quite scalable given that the user will be forced to set all sorts of parameters they don't care about. There are several approaches here:
  
  1. Have helper constructors with fewer parameters
  2. Have it a structure with `Default` implemented for it and multiple features behind options/enums as fields.
  3. Have distinct material types, at least on the user side, all implementing `Default`.

When designing this, the biggest question is how to properly support custom materials.

alteous:
### Proposed ideas

Both ideas revolve around making `Material` into a `struct`.

#### Non-opaque

```rust
pub enum Pipeline {
    Basic,
    Lambert,
    Phong,
    Pbr
    Sprite,
    // etc.
}

pub struct Material {
    pub pipeline: Pipeline,
    pub color: Color,
    pub map: Option<Texture<[f32; 4]>>,
    pub wireframe: bool,
    pub flat: bool,
    pub glossiness: f32.

    pub base_color_factor: [f32; 4],
    pub metallic_roughness: [f32; 2],
    pub occlusion_strength: f32,
    pub emissive_factor: [f32; 3],
    pub normal_scale: f32,
    pub base_color_map: Option<Texture<[f32; 4]>>,
    pub normal_map: Option<Texture<[f32; 4]>>,
    pub emissive_map: Option<Texture<[f32; 4]>>,
    pub metallic_roughness_map: Option<Texture<[f32; 4]>>,
    pub occlusion_map: Option<Texture<[f32; 4]>>,
}
```

#### Opaque

```rust
/* private */ enum Pipeline {
    Basic,
    Lambert,
    Phong,
    Pbr
    Sprite,
    // etc.
}

pub struct Material {
    pipeline: Pipeline,
    bits: u32,
    /// Corresponds to `color` in the basic pipeline and `base_color_factor` in the PBR pipeline etc.
    color: Color,
    params: [f32; 12],
    textures: [Option<Texture<[f32; 4]>>; 6],
}

impl Material {
    pub fn basic(color: Color, map: Option<Texture<[f32; 4]>, wireframe: bool) {
        Material {
            pipeline: Basic,
            color,
            textures: [map, None, None, None, None, None],
            bits: WIREFRAME_BIT,
        }
    }
}
```

vitvakatu:
I don't think we need feature flags here.

kvark:
Thanks for the detailed description!

[...]

PR #105, opened 2017-09-13
Andy

wallstbrok:
create builder patterns for material


alteous:
This is really helpful. Thank you!

vitvakatu:
I'll review in an hour or so

examples/materials.rs
```diff
@@ -14,23 +14,18 @@ fn main() {
...
-            occlusion_strength: 0.2,
-            emissive_factor: [0.0, 0.0, 0.0],
-            normal_scale: 1.0,
-            base_color_map: None,
-            normal_map: None,
-            emissive_map: None,
-            metallic_roughness_map: None,
-            occlusion_map: None,
-        },
+        three::material::MeshBasicBuilder::new().color(0xffffff).build(),
```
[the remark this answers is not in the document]
vitvakatu: I'm voting for `three::Material::mesh_basic`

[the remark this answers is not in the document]
alteous: > I'm voting for three::Material::mesh_basic

Actually, that's much better.

src/material/builder.rs
```diff
@@ -0,0 +1,203 @@
...
+        self
+    }
+
+    pub fn build(&self) -> Material{
+        Material::MeshLambert{
+            color: self.color,
+            flat: self.flat
+        }
+    }   
+
```
vitvakatu: nit: unneeded newline

src/material/builder.rs
```diff
@@ -0,0 +1,203 @@
...
+            emissive_factor: self.emissive_factor,
+            normal_scale: self.normal_scale,
+            base_color_map: self.base_color_map.clone(),
+            normal_map: self.normal_map.clone(),
+            emissive_map: self.emissive_map.clone(),
+            metallic_roughness_map: self.metallic_roughness_map.clone(),
+            occlusion_map: self.occlusion_map.clone()
+        }
+    }
+}
```
vitvakatu: nit: no newline at the end of the file

src/material/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    pub fn mesh_phong() -> builder::MeshPhong{
+        builder::MeshPhong::new()
+    }
+
+    pub fn mesh_pbr() -> builder::MeshPbr{
+        builder::MeshPbr::new()
+     }
+}
```
vitvakatu: nit: no newline at the end of the file

kvark:
Related - #31

examples/materials.rs
```diff
@@ -14,23 +14,18 @@ fn main() {
...
-            occlusion_strength: 0.2,
-            emissive_factor: [0.0, 0.0, 0.0],
-            normal_scale: 1.0,
-            base_color_map: None,
-            normal_map: None,
-            emissive_map: None,
-            metallic_roughness_map: None,
-            occlusion_map: None,
-        },
+        three::Material::mesh_basic().color(0xffffff).build(),
```
kvark: I think we should go deeper: user says `three::Material::mesh()` then builds some properties, and then `build` succeeds if Three is able to map those properties to enumeration types we have. E.g. one day we merge lambertian and basic shaders (just a non-realistic example, bear with me), and users will not need to do anything, since it's the `build` implementation logic that only going to change.

src/material/builder.rs
```diff
@@ -0,0 +1,203 @@
+use Material;
+use Texture;
+
+#[derive(Debug)]
+pub struct MeshBasic {
```
kvark: if we are introducing the structs, then build pattern is not really necessary, strictly speaking, i.e. a material struct can implement `Default` as well as `Into<MaterialEnum>` trait returning the same enumeration type that we can serialize, and the user then does something like:
```rust
let material = three::MeshBasic { color, .. default() }.into();
```

I haven't made my mind yet on which approach is better, open for opinions!

alteous:
Can bors format source code automatically before merging? I hate to dismiss PRs because of formatting issues.

src/material/builder.rs
alteous: I'm inclined to agree.

vitvakatu:
@alteous we can do that if we'll use `cargo-fmt`, I think.

kvark:
@alteous 
> Can bors format source code automatically before merging? I hate to dismiss PRs because of formatting issues.

I don't think we should make bors do anything about the source. Instead, we could have the rustfmt config here in the repo and just ask the PR author to run rustfmt once.

vitvakatu:
@kvark you raised a good point. I think we need RFC for new material design, it's not going to be so straitforward.

alteous:
Now that #128 has been merged, perhaps this PR could be updated to the new strategy? All the boilerplate code is there; it should just be a matter of renaming things.

vitvakatu:
Do we need builders now?

alteous:
It does seem a bit overkill now to include builders.

vitvakatu:
Agreed. I'm closing it now due to lack of response and not obvious profit

PR #106, opened 2017-09-13
Rustfmt pass

vitvakatu:
It's just an initial pass, give me your opinions about syntax.


src/factory/load_gltf.rs
```diff
@@ -82,41 +73,27 @@ impl super::Factory {
     }
 
     /// Loads a `glTF` material.
-    fn load_gltf_material(
-        &mut self,
-        mat: &gltf::Material,
-        buffers: &gltf_importer::Buffers,
-        base: &Path,
-    ) -> Material {
+    fn load_gltf_material(&mut self, mat: &gltf::Material, buffers: &gltf_importer::Buffers, base: &Path) -> Material {
```
kvark: I think the old formatting is better here

vitvakatu:
A pushed second commit with different option for function args. But it's not good.
It's weird behavior actually, default `Tall` value shouldn't provide these results.

vitvakatu:
If you have info how to fix that, let me know

vitvakatu:
Tomorrow, sorry. Going to sleep now

PR #108, opened 2017-09-16
Allow user to update positions of DynamicMesh

alteous:
This PR allows the user to alter the vertices of a `DynamicMesh` at runtime.

### DIFF
--- a/examples/mesh-update.rs
+++ b/examples/mesh-update.rs
@@ -1,0 +1,80 @@
+extern crate cgmath;
+extern crate three;
+extern crate mint;
+
+use cgmath::prelude::*;
+use std::f32::consts::PI;
+
+fn make_tetrahedron_geometry() -> three::Geometry {
+    let vertices = vec![
+        mint::Point3 {
+            x: 0.0,
+            y: 1.0,
+            z: 0.0,
+        },
+        mint::Point3 {
+            x: 0.0,
+            y: 0.0,
+            z: 1.0,
+        },
+        mint::Point3 {
+            x: (2.0 * PI / 3.0).sin(),
+            y: 0.0,
+            z: (2.0 * PI / 3.0).cos(),
+        },
+        mint::Point3 {
+            x: (4.0 * PI / 3.0).sin(),
+            y: 0.0,
+            z: (4.0 * PI / 3.0).cos(),
+        },
+    ];
+    let faces = vec![[0, 1, 2], [0, 2, 3], [0, 3, 1], [1, 3, 2]];
+    three::Geometry {
+        faces,
+        base_shape: three::geometry::Shape {
+            vertices,
+            ..three::geometry::Shape::empty()
+        },
+        ..three::Geometry::empty()
+    }
+}
+
+fn main() {
+    let shaders_path = concat!(env!("CARGO_MANIFEST_DIR"), "/data/shaders");
+    let title = "Three-rs Mesh Update Example";
+    let mut win = three::Window::builder(title, &shaders_path).build();
+    let cam = win.factory.perspective_camera(60.0, 1.0 .. 10.0);
+    let mut controls = three::controls::Orbit::builder(&cam)
+        .position([0.0, 2.0, -5.0])
+        .target([0.0, 0.0, 0.0])
+        .build();
+
+    let geometry = make_tetrahedron_geometry();
+    let material = three::Material::MeshBasic {
+        color: 0xFFFF00,
+        wireframe: true,
+        map: None,
+    };
+    let mut mesh = win.factory.mesh_dynamic(geometry, material);
+    let vertex_count = mesh.vertex_count();
+    win.scene.add(&mesh);
+
+    let mut timer = win.input.time();
+    let mut vi = 0;
+    while win.update() && !three::KEY_ESCAPE.is_hit(&win.input) {
+        let elapsed_time = timer.get(&win.input);
+        if elapsed_time > 1.0 {
+            // Reset the timer.
+            timer = win.input.time();
+            // Update the vertex `vi`.
+            let mut vmap = win.factory.map_vertices(&mut mesh);
+            let dir = cgmath::Vector4::from(vmap[vi].pos).truncate();
+            let pos = cgmath::Point3::from_vec(1.2 * dir);
+            vmap[vi].pos = [pos.x, pos.y, pos.z, 1.0];
+            // Increment vertex index.
+            vi = (vi + 1) % vertex_count;
+        }
+        controls.update(&win.input);
+        win.render(&cam);
+    }
+}
--- a/src/factory/mod.rs
+++ b/src/factory/mod.rs
@@ -65,6 +65,9 @@ const QUAD: [Vertex; 4] = [
     },
 ];
 
+/// Mapping writer.
+pub type MapVertices<'a> = gfx::mapping::Writer<'a, BackendResources, Vertex>;
+
 /// `Factory` is used to instantiate game objects.
 pub struct Factory {
     backend: BackendFactory,
@@ -275,12 +278,21 @@ impl Factory {
                 buffer: self.backend.create_index_buffer(data),
             }
         };
-        let (num_vertices, vertices) = {
+        let (num_vertices, vertices, upload_buf) = {
             let data = Self::mesh_vertices(&geometry.base_shape);
-            let buf = self.backend
+            let dest_buf = self.backend
                 .create_buffer_immutable(&data, gfx::buffer::Role::Vertex, gfx::memory::TRANSFER_DST)
                 .unwrap();
-            (data.len(), buf)
+            let upload_buf = self.backend.create_upload_buffer(data.len()).unwrap();
+            // TODO: Workaround for not having a 'write-to-slice' capability.
+            // Reason: The renderer copies the entire staging buffer upon updates.
+            {
+                self.backend
+                    .write_mapping(&upload_buf)
+                    .unwrap()
+                    .copy_from_slice(&data);
+            }
+            (data.len(), dest_buf, upload_buf)
         };
         let constants = self.backend.create_constant_buffer(1);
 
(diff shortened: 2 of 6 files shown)


src/mesh.rs
```diff
@@ -87,6 +88,16 @@ impl Mesh {
 }
 
 impl DynamicMesh {
+    /// Returns the indexed vertex position of the geometry base shape.
+    pub fn vertex(&self, index: usize) -> mint::Point3<f32> {
```
vitvakatu: Where is guarantee that we always can return value at any index? Need to use `Option` or smth else

src/factory/mod.rs
```diff
@@ -468,7 +485,44 @@ impl Factory {
...
+            TANGENT_X
+        };
+
+        let uv = if let Some(uv) = mesh.geometry.base_shape.tex_coords.get(index) {
+            uv.clone().into()
+        } else {
+            [0.0, 0.0]
+        };
+
+        mesh.geometry.base_shape.vertices[index] = position;
```
vitvakatu: index can be invalid, need to check

examples/mesh-update.rs
```diff
@@ -0,0 +1,167 @@
+extern crate cgmath;
```
vitvakatu: This example is too complicated. I would like to see either smaller one, or none at all.

vitvakatu:
By the way - it would be better to expose some way to access [`Writer`](https://docs.rs/gfx/0.16.1/gfx/mapping/struct.Writer.html) directly, because acquiring it for every single vertex isn't good.

src/mesh.rs
alteous: Agreed.

examples/mesh-update.rs
alteous: Yes, I agree. I'll re-implement with a tetrahedron instead of a cube.

src/factory/mod.rs
alteous: What would we do if the index is out of range?

alteous:
> it would be better to expose some way to access Writer directly

We have to be careful not to expose the internal `Vertex` type. How about the following?

```rust
pub fn overwrite_positions(
    &mut self,
    mesh: &mut DynamicMesh,
    updates: &[(usize, mint::Point3<f32>)],
);
```

vitvakatu:
Looks good as for me.

examples/mesh-update.rs
```diff
@@ -0,0 +1,75 @@
...
+    let mut controls = three::controls::Orbit::builder(&cam)
+        .position([0.0, 2.0, -5.0])
+        .target([0.0, 0.0, 0.0])
+        .build();
+
+    let mut dir_light = win.factory.directional_light(0xFFFFFF, 0.9);
+    dir_light.look_at([15.0, 35.0, 35.0], [0.0, 0.0, 2.0], None);
+    win.scene.add(&dir_light);
+
+    let geometry = make_tetrahedron_geometry();
```
alteous: Perhaps we should add a `Geometry::tetrahedron` constructor just for this example?

src/factory/mod.rs
```diff
@@ -468,7 +485,49 @@ impl Factory {
...
+    ///
+    /// Panics if any of the indices exceed the number of vertices in the
+    /// mesh geometry.
+    pub fn update_positions(
+        &mut self,
+        mesh: &mut DynamicMesh,
+        updates: &[(usize, mint::Point3<f32>)],
+    ) {
+        self.hub.lock().unwrap().update_mesh(mesh);
+        let mut mapping = self.backend.write_mapping(&mesh.dynamic.buffer).unwrap();
```
alteous: Only take one lock now.

src/factory/mod.rs
```diff
@@ -468,7 +485,49 @@ impl Factory {
...
+            };
+
+            let uv = if let Some(uv) = mesh.geometry.base_shape.tex_coords.get(index) {
+                uv.clone().into()
+            } else {
+                [0.0, 0.0]
+            };
+
+            mesh.geometry.base_shape.vertices[index] = position;
+            mapping[index] = Vertex {
```
alteous: The possibility of panic is noted in the docs.

src/factory/mod.rs
```diff
@@ -468,7 +485,49 @@ impl Factory {
...
     }
 
-    /// Update the geometry of `DynamicMesh`.
+    /// Update the vertices of the geometry of a `DynamicMesh`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if any of the indices exceed the number of vertices in the
+    /// mesh geometry.
+    pub fn update_positions(
```
alteous: Renamed from `overwrite_position`.

examples/mesh-update.rs
vitvakatu: Not sure, tetrahedron isn't best shape for everyday usage.
Example became much nicer, thanks!

src/mesh.rs
```diff
@@ -87,6 +88,17 @@ impl Mesh {
 }
 
 impl DynamicMesh {
+    /// Returns the indexed vertex position of the geometry base shape,
+    /// or `None` if the index is out of bounds.
+    pub fn vertex(&self, index: usize) -> Option<mint::Point3<f32>> {
+        self.geometry.base_shape.vertices.get(index).map(Clone::clone)
```
vitvakatu: `mint` types implement `Copy`, don't they?

alteous: Yes, but `Option` does not.

alteous:
Update: I've renamed the function `update_mesh_vertices` which now takes an `Iterator` instead of a slice.

alteous:
I would like to hear @kvark 's opinion too.

alteous:
Thinking ahead, suppose we have two separate functions `update_mesh_positions` and `update_mesh_normals`. If the user required both then we would have to map the buffer twice, which isn't ideal.

Perhaps later we should try another strategy. I'm thinking along the lines of.

```rust
factory
    .update_mesh()
    .write_vertices(iter::once((123, [0.0, 0.0, 0.0].into()))
    .write_normals(iter::once((456, [0.0, 1.0, 0.0].into())
    .exec();
```

src/mesh.rs
vitvakatu: Option has `cloned()` afaik, it looks much nicer

alteous: TIL, thanks. :smile:

examples/mesh-update.rs
```diff
@@ -0,0 +1,71 @@
+extern crate cgmath;
+extern crate three;
+extern crate mint;
+
+use cgmath::prelude::*;
+use std::iter;
+use std::f32::consts::PI;
+
+fn make_tetrahedron_geometry() -> three::Geometry {
+    let vertices = vec![
```
kvark: nice, I ❤️ tetrahedrons :)

src/factory/mod.rs
```diff
@@ -255,12 +255,29 @@ impl Factory {
...
+                .create_upload_buffer(data.len())
                 .unwrap();
-            (data.len(), buf)
+            // TODO: Workaround for not having a 'write-to-slice' capability.
+            // Reason: The renderer copies the entire staging buffer upon updates.
+            {
+                let mut mapping = self.backend
+                    .write_mapping(&upload_buf)
+                    .unwrap();
+                for (i, vertex) in data.iter().cloned().enumerate() {
```
kvark: how about this:
```rust
self.backend
    .write_mapping(&upload_buf)
    .unwrap()
    .copy_from_slice(&data);
```

src/factory/mod.rs
```diff
@@ -468,7 +485,26 @@ impl Factory {
...
     }
 
-    /// Update the geometry of `DynamicMesh`.
+    /// Update the vertex positions of the geometry of a `DynamicMesh`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if any of the indices exceed the number of vertices in the
+    /// mesh geometry.
+    pub fn update_mesh_vertices<I: Iterator<Item = (usize, mint::Point3<f32>)>>(
```
kvark: I think we should just provide a mapping API for dynamic meshes, e.g.
```rust
{
    let mut mapping = factory.map_vertices(&mesh);
    for i in 0.. 5 {
         mapping[i] = MyVertex {..};
    }
}
```
This mapping object would implement `DerefMut` to our vertex format and hold `factory` by a mutable reference, then issue an actual update message upon destruction (a message would make the renderer to copy the data from upload -> data buffer upon the first use or something).

src/mesh.rs
kvark: yep, let's just use `cloned` here

src/factory/mod.rs
alteous: This would be a much simpler approach. The downside is having to expose out vertex format.

alteous:
Should be ready to merge now.

src/factory/mod.rs
```diff
@@ -65,6 +65,9 @@ const QUAD: [Vertex; 4] = [
     },
 ];
 
+/// Mapping writer.
+pub type MapVertices<'a> = gfx::mapping::Writer<'a, BackendResources, Vertex>;
```
kvark: IIRC, the only place we currently expose gfx-rs is custom pipelines, which to be fair could be hidden behind a feature flag just fine. You proposed a wrapper before, I think that was a good idea, totally fine for a follow-up.

src/factory/mod.rs
```diff
@@ -275,12 +278,21 @@ impl Factory {
...
             let data = Self::mesh_vertices(&geometry.base_shape);
-            let buf = self.backend
+            let dest_buf = self.backend
                 .create_buffer_immutable(&data, gfx::buffer::Role::Vertex, gfx::memory::TRANSFER_DST)
                 .unwrap();
-            (data.len(), buf)
+            let upload_buf = self.backend.create_upload_buffer(data.len()).unwrap();
+            // TODO: Workaround for not having a 'write-to-slice' capability.
+            // Reason: The renderer copies the entire staging buffer upon updates.
+            {
```
kvark: I don't think you need the scope here?

PR #110, opened 2017-09-17
Added color pallets

csherratt:
Because programmers are bad at picking colours :P

It's a pointless change TBH, but some might find it useful.


src/scene.rs
```diff
@@ -39,3 +39,191 @@ impl Scene {
...
+    pub const GREEN: Color = 0x00_80_00;
+    pub const AQUA: Color = 0x00_ff_ff;
+    pub const TEAL: Color = 0x00_80_80;
+    pub const BLUE: Color = 0x00_00_ff;
+    pub const NAVY: Color = 0x00_00_80;
+    pub const FUCHSIA: Color = 0xff_00_ff;
+    pub const PURPLE: Color = 0x80_00_80;
+}
+
+/// This contains a list of crayon colors from a major crayon vendor.
```
alteous: > from a major crayon vendor.

:wink:

alteous:
I like the idea that of introducing a palette of colours that supposedly work together. The new and old colours ought to be in a separate `colors` module.

#### Concerns

* Legal issues regarding naming (I'm not a lawyer!)
* Might give a bad impression to new users. It would be a shame if new users thought `three` was only for really simple 3D graphics with solid colouring.

PR #113, opened 2017-09-18
Implement skybox

### LINKED ISSUE #102: Skybox

vitvakatu:
Closes #102

Also includes changes in texture background rendering.


data/shaders/skybox_vs.glsl
```diff
@@ -0,0 +1,21 @@
...
+
+void main() {
+    vec2 pos = gl_VertexID == 0 ? vec2(-1.0, -1.0) :
+               gl_VertexID == 1 ? vec2(-1.0,  1.0) :
+               gl_VertexID == 3 ? vec2( 1.0,  1.0) :
+                                  vec2( 1.0, -1.0) ;
+
+    vec4 a_Position = vec4(pos.xy, 1.0, 1.0);
+
+    mat4 inverseProj = inverse(u_Proj);
```
alteous: FYI, usually, matrix inverse operations are performed on the CPU. Since there are a tiny amount of vertices in the case of a skybox it doesn't really matter in this case.

examples/gltf.rs
```diff
@@ -21,6 +21,20 @@ fn main() {
         default
     };
 
+    // To enable skybox remove this if statement.
```
alteous: nit: `if` blocks are expressions in Rust.

src/texture.rs
```diff
@@ -69,3 +70,60 @@ impl<T> Texture<T> {
         ]
     }
 }
+
+/// Represents paths to cube map texture, useful for loading
+/// [`CubeMap`](struct.CubeMap.html).
+#[derive(Clone, Debug)]
+pub struct CubeMapPath<P: AsRef<Path>> {
```
alteous: Having a single `P: AsRef<Path>` type argument means every path source must be the same type.

```rust
let path = CubeMapPath {
    front: "front.png",
    back: Path::new("back.png"),
    // etc.
};
```
would fail to compile.

src/texture.rs
```diff
@@ -69,3 +70,60 @@ impl<T> Texture<T> {
...
+    /// "Right" image. `X+`.
+    pub right: P,
+    /// "Up" image. `Y+`.
+    pub up: P,
+    /// "Down" image. `Y-`.
+    pub down: P,
+}
+
+impl<P: AsRef<Path>> CubeMapPath<P> {
+    pub(crate) fn as_array(&self) -> [&P; 6] {
```
alteous: :+1:

alteous:
Nice choice of skybox textures.

data/shaders/skybox_vs.glsl
vitvakatu: It can be faster than on CPU due to only 4 vertices.

src/texture.rs
vitvakatu: I know. Do you have any proposal? I decided to do it this way because it's odd to use different path types for single cubemap

alteous: There are a few options:

1. Explicit type for every param
```rust
pub struct CubeMapPath<F, B, L, R, U, D>
where
    F: AsRef<Path>,
    B: AsRef<Path>,
    L: AsRef<Path>,
    R: AsRef<Path>,
    U: AsRef<Path>,
    D: AsRef<Path>,
{
     /// "Front" image. `Z+`.
    pub front: F,
    /// "Back" image. `Z-`.
    pub back: B,
    /// "Left" image. `X-`.
    pub left: L,
    /// "Right" image. `X+`.
    pub right: R,
    /// "Up" image. `Y+`.
    pub up: U,
    /// "Down" image. `Y-`.
    pub down: D,
}
```
2. Use dynamic dispatch.
```rust
pub struct CubeMapPath<'a> {
     /// "Front" image. `Z+`.
    pub front: &'a AsRef<Path>,
    /// "Back" image. `Z-`.
    pub back: &'a AsRef<Path>,
    /// "Left" image. `X-`.
    pub left: &'a AsRef<Path>,
    /// "Right" image. `X+`.
    pub right: &'a AsRef<Path>,
    /// "Up" image. `Y+`.
    pub up: &'a AsRef<Path>,
    /// "Down" image. `Y-`.
    pub down: &'a AsRef<Path>,
}
```
3. Use concrete type.
```rust
pub struct CubeMapPath<'a> {
     /// "Front" image. `Z+`.
    pub front: &'a Path,
    /// "Back" image. `Z-`.
    pub back: &'a Path,
    /// "Left" image. `X-`.
    pub left: &'a Path,
    /// "Right" image. `X+`.
    pub right: &'a Path,
    /// "Up" image. `Y+`.
    pub up: &'a Path,
    /// "Down" image. `Y-`.
    pub down: &'a Path,
}
```

alteous: Option 1 is zero-cost and most flexible, although ugly.

alteous:
https://github.com/three-rs/three/pull/113#discussion_r139405827 is my only concern.

src/texture.rs
alteous: You could try to return `[&Path; 6]` here instead.

src/texture.rs
vitvakatu: First one is too ugly (imagine if the user wants to store that inside some struct).
I don't know, I would like to avoid such complexity and have only one type, user can convert if he wants.

data/shaders/skybox_ps.glsl
```diff
@@ -0,0 +1,10 @@
+#version 150 core
+
+uniform samplerCube t_Cubemap;
+
+in vec3 v_TexCoord;
+out vec4 Target0;
+
+void main() {
+    Target0 = vec4(texture(t_Cubemap, v_TexCoord));
```
kvark: no need for `vec4` here, `texture` returns it anyway

data/shaders/skybox_vs.glsl
```diff
@@ -0,0 +1,21 @@
+#version 150 core
+#include globals
+
+out vec3 v_TexCoord;
+
+void main() {
+    vec2 pos = gl_VertexID == 0 ? vec2(-1.0, -1.0) :
```
kvark: You could take advantage of the vertex attributes of `quad_pipe` PSO if you use it

data/shaders/skybox_vs.glsl
kvark: Please don't do `inverse` in the shader. You added more stuff to `Globals` anyway, so better make it ready for computation on the CPU side. Inversing a matrix IIRC is not even available on non-GLSL shaders.

src/factory/mod.rs
```diff
@@ -575,15 +577,54 @@ impl Factory {
...
+        sampler: Sampler,
+        factory: &mut BackendFactory,
+    ) -> CubeMap<[f32; 4]> {
+        use gfx::texture as t;
+        let images = paths
+            .as_array()
+            .iter()
+            .map(|path| {
+                let format = Factory::parse_texture_format(path.as_ref());
+                let file = File::open(path).unwrap_or_else(|e| {
```
kvark: I think this is where `expect` should be used

src/factory/mod.rs
```diff
@@ -575,15 +577,54 @@ impl Factory {
...
+            &images[2],
+            &images[3],
+            &images[4],
+            &images[5],
+        ];
+        let size = images[0].dimensions().0;
+        let kind = t::Kind::Cube(size as t::Size);
+        let (_, view) = factory
+            .create_texture_immutable_u8::<gfx::format::Srgba8>(kind, &data)
+            .unwrap_or_else(|e| {
```
kvark: same here - please use `expect`

src/render.rs
```diff
@@ -211,7 +222,7 @@ where
     factory
         .link_program(code_vs.as_bytes(), code_ps.as_bytes())
         .map_err(|e| {
-            error!(
+            println!(
```
kvark: why this?

src/render.rs
```diff
@@ -144,6 +148,13 @@ gfx_defines! {
         color_target: gfx::RenderTarget<ColorFormat> = "Target0",
         depth_target: gfx::DepthTarget<DepthFormat> = gfx::preset::depth::LESS_EQUAL_WRITE,
     }
+
+    pipeline skybox_pipe {
```
kvark: I think you can re-use `quad_pipe` here?

src/render.rs
```diff
@@ -849,6 +857,52 @@ impl Renderer {
...
+                    params: self.quad_buf.clone(),
+                    resource: texture.to_param().0.raw().clone(),
+                    sampler: texture.to_param().1,
+                    target: self.out_color.clone(),
+                    depth_target: self.out_depth.clone(),
+                };
+                self.encoder.draw(&slice, &self.pso_quad, &data);
+            }
+            Background::Skybox(ref cubemap) => {
+                let slice = gfx::Slice {
```
kvark: looks like we can move this slice declaration up and share

src/texture.rs
kvark: I think it's fine for all the them to be same type here

src/render.rs
vitvakatu: Because I faced runtime errors a few times - and `error` doesn't print anything while `println` does. But yep, it was temporal change just for me, I'll revert

src/render.rs
vitvakatu: Sounds reasonable

src/render.rs
vitvakatu: I'll try

src/factory/mod.rs
vitvakatu: Why? Expect will not use error in the message.

kvark: Hmm, true.

data/shaders/skybox_vs.glsl
vitvakatu: I decided to leave it as it is. Quad pso doesn't have vertex coordinates, it would be overcomplication.

vitvakatu:
Ouch, rustfmt broke `gfx_defines`. Should I revert?

data/shaders/globals.glsl
```diff
@@ -0,0 +1,6 @@
+uniform b_Globals {
+    mat4 u_ViewProj;
+    mat4 u_InverseProj;
```
kvark: the name is incorrect. Should be something like `u_InverseRotateProj` to reflect the following facts:
  1. it does include the view transform
  2. but not the translation portion of it...

vitvakatu: Actually it's inverted camera projection matrix, it's not View matrix.

kvark: It's not **just** the inverted projection, that's my concern

PR #119, opened 2017-09-24
Expose underlaying glutin::GlWindow

### LINKED ISSUE #94: Access underlaying Glutin window
In some cases it might be useful (highly, even, in the case of conrod) to access the raw glutin window. Maybe a method that returns a pointer to it and the event loop?

vitvakatu:
Closes #94


vitvakatu:
It's behind feature flag, should be safe?

Issue #93, opened 2017-08-29
Panic on Aviator example (AMD)


vitvakatu:
```
PS D:\Projects\git\rust\three-rs> cargo run --example=aviator
    Finished dev [unoptimized + debuginfo] target(s) in 0.1 secs
     Running `target\debug\examples\aviator.exe`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DescriptorInit(ConstantBuffer("b_PbrParams", Some(Offset { name: "u_NormalScale", shader_offset: 64, code_offset: 56 })))', src\libcore\result.rs:860:4
stack backtrace:
   0: std::sys_common::backtrace::_print
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::result::unwrap_failed
             at C:\projects\rust\src\libcore/macros.rs:31
   9: <core::result::Result<T, E>>::unwrap
             at C:\projects\rust\src\libcore/result.rs:738
  10: three::render::Renderer::new
             at src/render.rs:346
  11: three::window::WindowBuilder::build
             at src/window.rs:76
  12: aviator::main
             at examples/aviator/main.rs:27
  13: _rust_maybe_catch_panic
  14: std::rt::lang_start
  15: main
  16: _tmainCRTStartup
  17: mainCRTStartup
  18: unit_addrs_search
error: process didn't exit successfully: `target\debug\examples\aviator.exe` (exit code: 101)
```

Video card: AMD Radeon R9 M275X

alteous:
Caused by insufficient padding on the `PbrParams::normal_scale` (`name: "u_NormalScale", shader_offset: 64, code_offset: 56`)

alteous:
I'm not sure how this could be fixed. @kvark has more experience with `gfx`, so perhaps he could suggest a more portable solution.

Is there an OpenGL extension that standardises shader variable offsets?

shmutalov:
But, it worked in previous versions :( Something was changed?

alteous:
[Std140](https://stackoverflow.com/questions/16270846/when-to-use-std140-in-opengl), for example?

alteous:
@shmutalov That's definitely weird. I've checked `git blame` but all seems unchanged. How long ago was it working?

[...]

PR #120, opened 2017-09-25
Use layout(std140)

alteous:
This PR adds standardises the layout of uniform blocks with `layout(std140)` to prevent hardware-dependent data alignment/padding issues. Fixes https://github.com/three-rs/three/issues/93.


alteous:
Requested a review from @kvark to check this change doesn't have any unwanted side effects.

kvark:
Nope, shipit!
Thanks :)
bors r+

PR #121, opened 2017-09-25
Add examples in geometry documentation

alteous:
Adds runnable code examples to the geometry module documentation.

This also renames `Geometry::sphere` as `Geometry::uv_sphere` to make way for icosahedral spheres.

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -6,7 +6,7 @@ use mint;
 use std::collections::HashMap;
 
 /// A shape of geometry that is used for mesh blending.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct Shape {
     /// Vertices.
     pub vertices: Vec<mint::Point3<f32>>,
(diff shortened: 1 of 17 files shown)


src/geometry.rs
```diff
@@ -19,20 +19,50 @@ pub struct Shape {
 }
 
 impl Shape {
-    /// Create an empty shape.
+    /// Creates an empty shape.
     pub fn empty() -> Self {
```
vitvakatu: Hmm... do we really need that method? One can use `Shape::default` now

src/geometry.rs
```diff
@@ -43,16 +73,57 @@ pub struct Geometry {
...
+    ///        base_shape: three::geometry::Shape {
+    ///            vertices,
+    ///            .. three::geometry::Shape::empty()
+    ///        },
+    ///        .. three::Geometry::empty()
+    ///    }
+    /// }
+    /// # fn main() { let _ = make_triangle(); }
+    /// ```
     pub fn empty() -> Self {
```
vitvakatu: Same here

src/geometry.rs
```diff
@@ -91,33 +162,77 @@ impl Geometry {
...
+    /// Unit square in the XY plane, centered at the origin.
+    ///
+    /// ```rust
+    /// # extern crate three;
+    /// fn make_square() -> three::Geometry {
+    ///     three::Geometry::plane(1.0, 1.0)
+    /// }
+    /// # fn main() { let _ = make_square(); }
+    /// ```
+    pub fn plane(width: f32, height: f32) -> Self {
```
vitvakatu: Did you run rustfmt?

alteous: No, actually I undid the formatting for this case.

src/geometry.rs
alteous: I'm inclined to agree. I kept this because there's a note in the Rust API guidelines about constructors calling `Default::default`. See [the bottom of this page](https://rust-lang-nursery.github.io/api-guidelines/predictability.html#constructors-are-static-inherent-methods-c-ctor).

alteous: > "Note that it is common and expected for types to implement both Default and an empty new constructor. new is the constructor convention in Rust, and users expect it to exist, so if it is reasonable for the basic constructor to take no arguments, then it should, even if it is functionally identical to default."

Perhaps we should rename it as `new` instead?

src/geometry.rs
vitvakatu: Either we need to run rustfmt every time, or never run it.

src/geometry.rs
alteous: > Some types' constructors, most notably I/O resource types, use distinct naming conventions for their constructors, as in File::open, Mmap::open, TcpStream::connect, and UpdSocket::bind. In these cases names are chosen as appropriate for the domain.

Actually, `empty` is more appropriate for our case.

vitvakatu: Well, let's keep `empty` as it is

src/geometry.rs
alteous: Fair enough.

alteous: I think [`array-width`](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#array_width) is not working correctly. It is set to 60 by default.

vitvakatu: Yep, looks like

vitvakatu:
@alteous did you change rustfmt config? I see multiple changes in code style, but config wasn't updated.

alteous:
No, the configuration is unchanged. I ran `cargo fmt` (version 0.21.0). I have `rustfmt` version 0.9.0-nightly installed.

vitvakatu:
Okay then

vitvakatu:
There are no breaking changes in this PR, so
bors r+

alteous:
@vitvakatu Are you happy for this to be merged?

PR #123, opened 2017-09-27
[WIP] Shader code system

alteous:
This is an attempt to redesign the shader code system, as discussed in issue #28.

Currently, the default shaders have been baked into the program. See `build.rs` and `src/data.rs` to find out how this was achieved. There is no way of overriding shaders yet, but one can build custom pipelines as per usual.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -13,7 +13,7 @@ use mint;
 
 use std::collections::HashMap;
 use std::mem;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
 pub use self::back::CommandBuffer as BackendCommandBuffer;
 pub use self::back::Factory as BackendFactory;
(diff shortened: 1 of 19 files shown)


src/render.rs
```diff
@@ -45,6 +45,94 @@ const STENCIL_SIDE: gfx::state::StencilSide = gfx::state::StencilSide {
...
+                "skybox",
+                "sprite",
+            ];
+            for name in &shader_names {
+                use std::borrow::Borrow;
+                let vs = data::SHADERS.get(&format!("data/shaders/{}_vs.glsl", name)).unwrap();
+                let fs = data::SHADERS.get(&format!("data/shaders/{}_ps.glsl", name)).unwrap();
+                let ss = ShaderSource {
+                    name: name.to_string(),
+                    vertex: manager.preprocess(str::from_utf8(vs.borrow()).unwrap()),
```
alteous: To avoid deadlock.

src/render.rs
```diff
@@ -45,6 +45,94 @@ const STENCIL_SIDE: gfx::state::StencilSide = gfx::state::StencilSide {
...
+    }
+
+    impl ShaderSourceManager {
+        fn preprocess(&self, code: &str) -> String {
+            let mut new_code = String::new();
+            for line in code.lines() {
+                if line.starts_with("#include") {
+                    for dep_name in line.split(' ').skip(1) {
+                        match dep_name {
+                            "locals" => new_code += &self.locals,
```
alteous: If these change then we have to re-compile the whole program anyway.

src/window.rs
```diff
@@ -113,16 +109,17 @@ impl Builder {
 }
 
 impl Window {
+    /// Create a new window with default parameters.
+    pub fn new<T: Into<String>>(title: T) -> Self {
```
alteous: The window creation interface is greatly simplified for simple cases.

alteous:
Actually, this is overly complex. I'll redesign.

PR #124, opened 2017-09-27
Add shader code (re)loading

alteous:
This PR prepares for a more sophisticated shader system as discussed in #28.

The window creation API is greatly simplified for simple use cases.


src/window.rs
```diff
@@ -113,16 +109,17 @@ impl Builder {
 }
 
 impl Window {
+    /// Create a new window with default parameters.
+    pub fn new<T: Into<String>>(title: T) -> Self {
```
vitvakatu: 👍

src/render.rs
```diff
@@ -296,6 +226,206 @@ struct DebugQuad {
...
+
+    /// Corresponds to `Material::MeshPbr`.
+    pbr: gfx::PipelineState<back::Resources, pbr_pipe::Meta>,
+
+    /// Used internally for rendering `Background::Skybox`.
+    skybox: gfx::PipelineState<back::Resources, quad_pipe::Meta>,
+}
+
+impl PipelineStates {
+    pub fn init(src: &ProgramSourceSet, factory: &mut back::Factory) -> Self {
```
alteous: The idea here is if the `ProgramSourceSet` needs to change then we can reload *all* the shaders in one go.

src/render.rs
```diff
@@ -331,35 +452,104 @@ pub struct Renderer {
...
 impl Renderer {
     #[cfg(feature = "opengl")]
-    #[doc(hidden)]
-    pub fn new(
+    pub(crate) fn new(
         builder: glutin::WindowBuilder,
         context: glutin::ContextBuilder,
         event_loop: &glutin::EventsLoop,
-        shader_path: &Path,
+        source: &ProgramSourceSet,
```
alteous: Preparing shaders is now delegated to the caller.

build.rs
```diff
@@ -0,0 +1,10 @@
+extern crate includedir_codegen;
+
+use includedir_codegen::Compression;
+
+fn main() {
+    includedir_codegen::start("FILES")
+        .dir("data/shaders", Compression::None)
```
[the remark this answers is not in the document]
alteous: It's so little data I didn't think it would be worth it. Would gzip be adequate?

alteous:
I forgot about the overriding scheme actually! Will fix in the next commit. Auto-reload will be delegated to the user.

src/render.rs
```diff
@@ -331,35 +452,104 @@ pub struct Renderer {
...
+    /// Load the named shader from the default set of shaders.
+    pub fn default(name: &str, suffix: &str) -> io::Result<Self> {
+        let path = format!("data/shaders/{}_{}.glsl", name, suffix);
+        let unprocessed = data::FILES.get(&path).unwrap();
+        let processed = Self::preprocess("", str::from_utf8(unprocessed.borrow()).unwrap())?;
+        Ok(ShaderSource { code: processed })
+    }
+
+    /// Load the named shader from the given directory path.
+    pub fn user(root: &str, name: &str, suffix: &str) -> io::Result<Self> {
```
vitvakatu: I would like to see `&Path`s here, not `&str`.

alteous:
I implore you to try out the `reload` example! :smile:

alteous:
The shader code was adapted from http://nuclear.mutantstargoat.com/articles/sdr_fract/ (public domain).

alteous:
Should be ready to merge now.

examples/reload.rs
```diff
@@ -0,0 +1,136 @@
...
+            let source_set = three::render::source::Set {
+                sprite: three::render::source::Sprite::user(&dir).unwrap(),
+                ..Default::default()
+            };
+            match three::render::PipelineStates::new(&source_set, &mut win.factory) {
+                Ok(pipeline_states) => {
+                    win.renderer.reload(pipeline_states);
+                    reload = false;
+                }
+                Err(err) => println!("{:#?}", err),
```
kvark: leaving `reload==true` would mean the attempts to reload happen each successive frame, accompanied by the `println` spam. I suppose we need to reset `reload` flag regardless?

src/render/mod.rs.bk
```diff
@@ -0,0 +1,957 @@
+use cgmath::{Matrix4, SquareMatrix, Transform as Transform_, Vector3};
```
kvark: this file does not belong here

src/render/source.rs
```diff
@@ -0,0 +1,133 @@
...
+        root: P,
+        code: &str,
+    ) -> io::Result<String> {
+        let root = root.as_ref();
+        let mut new_code = String::new();
+        for line in code.lines() {
+            if line.starts_with("#include") {
+                for dep_name in line.split(' ').skip(1) {
+                    match dep_name {
+                        "locals" | "lights" | "globals" => {
```
kvark: hmm, I'd prefer us to have a solid distinction of the built-in versus custom includes in shaders
something like `locals` for our shaders and `/locals` for user shaders

src/window.rs
```diff
@@ -90,16 +98,58 @@ impl Builder {
...
         let context = glutin::ContextBuilder::new()
             .with_vsync(self.vsync)
             .with_multisampling(self.multisampling);
 
+        let mut source_set = render::source::Set::default();
+        if let Some(path) = self.shader_directory.as_ref() {
+            let path = path.to_str().unwrap();
+            macro_rules! try_override {
+                ($name:ident) => {
+                    if let Ok(src) = render::Source::user(
```
kvark: nit: let's use `match` against `if let ... else` here and below

examples/reload.rs
alteous: That's an oversight, thanks.

src/window.rs
alteous: Much better idea.

src/render/source.rs
alteous: Is that not covered already? :confused:

alteous:
Hooray! :fireworks:

PR #126, opened 2017-10-06
[WIP] Input rework

vitvakatu:
Don't worry about formatting for now - I need to update rustfmt.

Still WIP, needs documentationa and a few more questions resolving.
I'm open to suggestions, as always.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,9 @@ cache: cargo
 rust:
   - nightly
   - stable
-
+branches:
+  except:
+    - staging.tmp
 
 # Install ALSA development libraries before compiling on Linux.
 addons:
--- a/README.md
+++ b/README.md
@@ -15,22 +15,20 @@ If you a looking for something simple to prototype 3D graphics with, you found i
 
 ## Motivation and Goals
 
-  1. Ergonomics is first priority. Being able to prototype quickly and code intuitively is more important than capturing all the 3D features. We already have a solid foundation with [gfx-rs](https://github.com/gfx-rs/gfx), so let's make some use of it by providing a nice higher-level abstraction.
-  2. Follow "Three.JS" - this is simpler than coming up with a brand new API (like [kiss3d](https://github.com/sebcrozet/kiss3d)), yet instantly familiar to a large portion of Web develper. Some deviations from the original API are imminent.
-  3. Explore the capabilities of Rust to mimic deep object-oriented nature of JavaScript. This is a tough challenge, involving channels, defer implementations, blood, and sweat.
-  4. Battle-test the [genmesh](https://github.com/gfx-rs/genmesh) library. Being able to create cubes, spheres, cylinders (and more) with one-liners allows for nice proceduraly generated demos.
-  5. Play with Component-Graph System concept, provided by [froggy](https://github.com/kvark/froggy). It's a good fit for the scene graph implementation that is fast and usable.
+Three-rs is focused at the ergonomics first and foremost. Being able to prototype quickly and code intuitively is more important than capturing all the 3D features. We already have a solid foundation with [gfx-rs](https://github.com/gfx-rs/gfx), so let's make some use of it by providing a nice higher-level abstraction.
+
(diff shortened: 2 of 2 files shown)


examples/anim.rs
```diff
@@ -136,7 +136,7 @@ fn main() {
     let blend_time = 0.2f32;
     let mut timer = win.input.time();
 
-    while win.update() && !three::KEY_ESCAPE.is_hit(&win.input) {
+    while win.update() && !win.input.button_down(three::KEY_ESCAPE) {
```
kvark: careful with semantics! What this should be doing is checking if `Esc` was hit *at some point* between the last frame and this one. This is not the same as checking for the `Esc` to be *currently* pressed, which is what `button_down` implies to me.

src/controls/first_person.rs
```diff
@@ -3,25 +3,25 @@ use mint;
...
+use input::{Axis, Input, Key};
 use object::Object;
 use std::f32::consts::PI;
 
 #[derive(Clone, Debug)]
 struct Axes {
-    pub forward: Option<KeyAxis>,
-    pub strafing: Option<KeyAxis>,
-    pub vertical: Option<KeyAxis>,
+    pub forward: Option<Axis>,
```
kvark: The reason I made `KeyAxis` a separate type as opposed to a more generic `Axis` is because with keys it's discreet, while mouse/joystick axis is continuous. This semantics difference makes for different APIs making sense, thus - different types.

src/input.rs
```diff
@@ -203,6 +225,106 @@ impl Input {
...
+                }
+                if pos_count + neg_count == 0 {
+                    None
+                } else {
+                    Some(pos_count + neg_count)
+                }
+            }
+            Axis::Custom(axis_id) => {
+                let mut count = 0i8;
+                for &axis in self.1.axes_raw.iter() {
```
kvark: could use `filter().count()`

src/input.rs
```diff
@@ -203,6 +225,106 @@ impl Input {
...
+                    None
+                }
+            }
+            Axis::Custom(axis_id) => {
+                let moves = self.1
+                    .axes_raw
+                    .iter()
+                    .filter(|&&(id, _)| id == axis_id)
+                    .map(|&(_, value)| value)
+                    .collect::<Vec<_>>();
```
kvark: strictly speaking, this doesn't need to be collected into a `Vec` since all you need is to iterate it here for the `sum`

src/controls/first_person.rs
alteous: Perhaps we should rename to make this distinction clear, i.e. `Discrete` and `Continuous` axes.

examples/anim.rs
vitvakatu: Then `is_hit` is a bad naming too. Perhaps we can use `button_pressed`?

src/controls/first_person.rs
vitvakatu: I thought we've reached consensus... I'm not sure. I need to check one thing for possibility of implementation before final decision...

src/input.rs
vitvakatu: Thanks!

src/input.rs
vitvakatu: I know, but I need to return `None` in case of no elements in this vector. I'm not sure is it possible without `collect`

kvark: You could do something like
```rust
let iterator = ...;
if iterator.clone().count() != 0 {
   iterator.sum()
} else {
  None
}
```

src/controls/first_person.rs
kvark: > Accessing input data with indexing (like "input.pressed()[button]") is impossible due to current implementation of ops::Index and borrow checker

Is it, really? Please provide what you got. I think `input[button].is_hit()` would be quite possible, assuming `button` is `Copy`, and `input` stores actual structures per key that it can return as references.

vitvakatu: Well, your syntax (`input[button].is_hit()`) is possible, yes. Impossible one is `input.pressed()[button]`

alteous: The former is more intuitive.

PR #128, opened 2017-10-10
New material strategy

alteous:
In response to #105 , this PR changes the way materials are specified.

* There are explicit data structures for each of the pipelines.
~~* The `LineBasic` material and its implementation has been removed.~~
~~* Hence, only mesh materials will be supported.~~
* There are explicit pipelines `Basic`, `basic::Custom`, and `Wireframe`.

### DIFF
--- a/examples/shapes.rs
+++ b/examples/shapes.rs
@@ -11,7 +11,7 @@ fn main() {
 
     let mut mbox = {
         let geometry = three::Geometry::cuboid(3.0, 2.0, 1.0);
-        let material = three::material::Basic { color: 0x00FF00, map: None };
+        let material = three::material::Wireframe { color: 0x00FF00 };
         win.factory.mesh(geometry, material)
     };
     mbox.set_position([-3.0, -3.0, 0.0]);
@@ -19,7 +19,7 @@ fn main() {
 
     let mut mcyl = {
         let geometry = three::Geometry::cylinder(1.0, 2.0, 2.0, 5);
-        let material = three::material::Basic { color: 0xFF0000, map: None };
+        let material = three::material::Wireframe { color: 0xFF0000 };
         win.factory.mesh(geometry, material)
     };
     mcyl.set_position([3.0, -3.0, 0.0]);
@@ -27,7 +27,7 @@ fn main() {
 
     let mut msphere = {
         let geometry = three::Geometry::uv_sphere(2.0, 5, 5);
-        let material = three::material::Basic { color: 0xFF0000, map: None };
+        let material = three::material::Wireframe { color: 0xFF0000 };
         win.factory.mesh(geometry, material)
     };
     msphere.set_position([-3.0, 3.0, 0.0]);
--- a/examples/shapes.rs
+++ b/examples/shapes.rs
@@ -11,11 +11,7 @@ fn main() {
 
     let mut mbox = {
         let geometry = three::Geometry::cuboid(3.0, 2.0, 1.0);
-        let material = three::material::Basic {
-            color: 0x00ff00,
-            map: None,
-            pipeline: three::material::basic::Pipeline::Wireframe,
-        };
+        let material = three::material::Basic { color: 0x00FF00, map: None };
         win.factory.mesh(geometry, material)
     };
     mbox.set_position([-3.0, -3.0, 0.0]);
@@ -23,11 +19,7 @@ fn main() {
 
     let mut mcyl = {
         let geometry = three::Geometry::cylinder(1.0, 2.0, 2.0, 5);
-        let material = three::material::Basic {
-            color: 0xff0000,
-            map: None,
-            pipeline: three::material::basic::Pipeline::Wireframe,
-        };
+        let material = three::material::Basic { color: 0xFF0000, map: None };
         win.factory.mesh(geometry, material)
     };
     mcyl.set_position([3.0, -3.0, 0.0]);
@@ -35,15 +27,23 @@ fn main() {
 
     let mut msphere = {
         let geometry = three::Geometry::uv_sphere(2.0, 5, 5);
-        let material = three::material::Basic {
-            color: 0xff0000,
-            map: None,
-            pipeline: three::material::basic::Pipeline::Wireframe,
-        };
+        let material = three::material::Basic { color: 0xFF0000, map: None };
         win.factory.mesh(geometry, material)
     };
     msphere.set_position([-3.0, 3.0, 0.0]);
     win.scene.add(&msphere);
+
+    let mut mline = {
+        let geometry = three::Geometry::with_vertices(vec![
+            [-2.0, -1.0, 0.0].into(),
+            [0.0, 1.0, 0.0].into(),
+            [2.0, -1.0, 0.0].into(),
+        ]);
+        let material = three::material::Line { color: 0x0000FF };
+        win.factory.mesh(geometry, material)
+    };
+    mline.set_position([3.0, 3.0, 0.0]);
+    win.scene.add(&mline);
 
     let mut angle = cgmath::Rad::zero();
     while win.update() && !three::KEY_ESCAPE.is_hit(&win.input) {
@@ -53,8 +53,8 @@ fn main() {
             mbox.set_orientation(q);
             mcyl.set_orientation(q);
             msphere.set_orientation(q);
+            mline.set_orientation(q);
         }
-
         win.render(&cam);
     }
 }
--- a/examples/shapes.rs
+++ b/examples/shapes.rs
@@ -11,10 +11,10 @@ fn main() {
 
     let mut mbox = {
         let geometry = three::Geometry::cuboid(3.0, 2.0, 1.0);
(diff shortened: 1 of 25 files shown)


src/material.rs
```diff
@@ -1,45 +1,220 @@
...
+        ///
+        /// Default: `Solid`.
+        pub pipeline: Pipeline,
+    }
+
+    /// Specifies which pipeline should be used to render the mesh.
+    #[derive(Clone, Debug)]
+    pub enum Pipeline {
+        /// Renders the mesh as a solid.
+        Solid,
```
alteous: Perhaps this should be renamed `Fill` to mimic the implementation pipeline (`pso_mesh_basic_fill`)?

src/material.rs
```diff
@@ -1,45 +1,220 @@
...
+
+/// Parameters for a Lamberian diffusion reflection model.
+#[derive(Clone, Debug, Default)]
+pub struct Lambert {
+    /// Solid color applied in the absense of `map`.
+    ///
+    /// Default: `0x000000` (black).
+    pub color: Color,
+
+    /// Specifies whether lighting should be constant over faces.
```
alteous: Please correct me if this is inaccurate. I'm unfamiliar with the Lambert model.

src/material.rs
```diff
@@ -1,45 +1,220 @@
...
+    /// Default: `false` (lighting is interpolated across faces).
+    pub flat: bool,
+}
+
+/// Parameters for a PBR (physically based rendering) lighting model.
+#[derive(Clone, Debug)]
+pub struct Pbr {
+    /// Solid base color applied in the absense of `base_color_map`.
+    ///
+    /// Default: `0xFFFFFF` (white).
```
alteous: All other defaults are black. Should we unify, i.e. make all 'base colours' white by default?

src/material.rs
```diff
@@ -1,45 +1,220 @@
...
-    },
-    /// 2D Sprite.
-    Sprite { map: Texture<[f32; 4]> },
-    /// Custom material.
-    CustomBasicPipeline {
-        color: Color,
-        map: Option<Texture<[f32; 4]>>,
-        pipeline: BasicPipelineState,
-    },
+pub struct Material(pub(crate) Params);
```
alteous: I think it's important for `Material` to be a `struct` since it's more accessible in the top-level documentation.

src/render/mod.rs
```diff
@@ -789,31 +780,36 @@ impl Renderer {
...
-                        } => (&self.pso.mesh_basic_fill, color, 0.0, map.as_ref()),
-                        Material::MeshBasic {
-                            color,
-                            map: _,
-                            wireframe: true,
-                        } => (&self.pso.mesh_basic_wireframe, color, 0.0, None),
-                        Material::MeshLambert { color, flat } => (
+                        Params::Pbr(_) => unreachable!(),
+                        Params::Basic(ref params) => {
+                            match params.pipeline {
```
vitvakatu: tip: Code size can be reduced by moving match inside created tuple

alteous: This is what `rustfmt` chose, I think. :frowning:

src/color.rs
```diff
@@ -0,0 +1,65 @@
...
+            x / 12.92
+        }
+    };
+    [f(c >> 16), f(c >> 8), f(c)]
+}
+
+/// Linear to sRGB conversion.
+///
+/// Implementation taken from https://en.wikipedia.org/wiki/SRGB
+pub fn from_linear_rgb(c: [f32; 3]) -> Color {
```
alteous: Perhaps `fn from_linear_rgb(r: f32, g: f32, b: f32)` instead?

src/factory/mod.rs
```diff
@@ -325,20 +325,38 @@ impl Factory {
...
             _ => unreachable!(),
-        });
-        Mesh { object: hub.spawn_visual(mat, gpu_data) }
+        };
+        Mesh { object: hub.spawn_visual(material, gpu_data) }
+    }
+
+    /// Create a `Mesh` sharing the geometry with another one but with a different material.
+    /// Rendering a sequence of meshes with the same geometry is faster.
+    pub fn mesh_instance_with_material<M: Into<Material>>(
```
alteous: This replaces the previous `Option<Material>` idea. I wasn't sure about how to handle the type parameter `fn mesh_instance<M: Into<Material>>(&mut self, template: &Mesh, material: Option<M>)` when `material` is `None`.

src/factory/mod.rs
```diff
@@ -689,34 +707,38 @@ impl Factory {
...
-                wireframe: false,
-            },
-            _ => Material::MeshBasic {
-                color: 0xffffff,
-                map: None,
-                wireframe: true,
-            },
+            } => {
+                material::Basic {
+                    color: cf2u(color),
```
alteous: Do we need `cf2u` anymore?

src/material.rs
```diff
@@ -1,45 +1,220 @@
...
+pub struct Sprite {
+    /// The texture the apply to the sprite.
+    pub map: Texture<[f32; 4]>,
+}
+
+/// Specifies the appearance of a [`Mesh`](struct.Mesh.html).
 #[derive(Clone, Debug)]
-pub enum Material {
-    /// Basic wireframe with specific `Color`.
-    LineBasic { color: Color },
```
alteous: Bye bye!

src/color.rs
```diff
@@ -0,0 +1,65 @@
+//! sRGB colors.
```
kvark: nice module!

examples/shapes.rs
```diff
@@ -35,36 +35,23 @@ fn main() {
...
-    win.scene.add(&mline);
-
     let mut angle = cgmath::Rad::zero();
     while win.update() && !three::KEY_ESCAPE.is_hit(&win.input) {
         if let Some(diff) = three::AXIS_LEFT_RIGHT.timed(&win.input) {
             angle += cgmath::Rad(1.5 * diff);
             let q = cgmath::Quaternion::from_angle_y(angle);
             mbox.set_orientation(q);
             mcyl.set_orientation(q);
-            mline.set_orientation(q);
```
kvark: why is line removed?

src/material.rs
```diff
@@ -1,45 +1,220 @@
...
-#[allow(missing_docs)]
+pub use self::basic::Basic;
+
+/// Basic material parameters.
+pub mod basic {
+    use super::*;
+
+    /// Parameters for a basic mesh material.
+    #[derive(Clone, Debug, Default)]
+    pub struct Basic {
```
kvark: I feel that `basic` module serves little purpose. Would be easier to just move `Pipeline` up

src/material.rs
kvark: What's wrong with `enum` accessibility?

examples/shapes.rs
alteous: Because support for rendering lines is removed.

kvark: yes, but why?

src/material.rs
vitvakatu: Enums are placed in the separate section in docs.

examples/shapes.rs
alteous: I don't feel it belongs in the API. The lines are not anti-aliased, they make the pipelines more complex, they don't match up at the ends, and they're seldom even used. I think it would be a better option to provide a line geometry constructor/builder or an explicit `LineMesh` type instead, although I'm not going to include that in this PR.

alteous: Relevant: https://blog.mapbox.com/drawing-antialiased-lines-with-opengl-8766f34192dc

alteous: I could add it back but rename the material just `Line` instead of `LineBasic`like we have the `Sprite` material.

kvark: I don't find the arguments convincing. Lines are not ideal, but they do allow for user cases previously inaccessible, and they still use the same PSO definition so not much needed to support from our side. Please bring `Line` back in some form.

alteous: Okay, will do. :smile:

src/material.rs
alteous: The idea is to put the basic pipeline in its own namespace so we can add others without them conflicting. I'll move it up for now.

src/material.rs
alteous: ^^ :+1:

alteous: It's a minor concern though. I'll put it back for now since there's not much justification for an opaque material type right now.

src/material.rs
alteous: I've removed `Pipeline` in favour of three explicit materials `Basic`, `basic::Custom`, and `Wireframe`.

src/material.rs
alteous: ^^ I decided to make all base colours `WHITE` by default.

src/material.rs
alteous: I've made it an enum again for simplicity. We can always revisit this.

PR #130, opened 2017-10-12
Input rework

vitvakatu:


alteous:
What is the meaning of `win.input.button(three::KEY_ESCAPE).released` now? I take 'released' to mean 'the key was down and now it is up'.

vitvakatu:
Yep, it was one of concerns. It's the opposite of `pressed`, so I can remove it and see `!input.button(ESCAPE).pressed` instead. Is it ok?

alteous:
That would be more correct.

examples/anim.rs
```diff
@@ -134,7 +134,7 @@ fn main() {
     let blend_time = 0.2f32;
     let mut timer = win.input.time();
 
-    while win.update() && !three::KEY_ESCAPE.is_hit(&win.input) {
+    while win.update() && !win.input.button(three::KEY_ESCAPE).pressed {
```
kvark: hmm, that worries me. The new code is more verbose, while we need to make it shorter

vitvakatu: From my point of view, code is little more verbose but much clearer now. If we're talking about amount of symbols, I don't see any way to reduce it atm.

src/input/mod.rs
```diff
@@ -26,7 +26,7 @@ struct State {
     mouse_pos_ndc: mint::Point2<f32>,
 }
 
-struct Delta {
+struct Diff {
```
alteous: We've gone back to diff?

src/input/mod.rs
```diff
@@ -235,128 +235,148 @@ impl Input {
...
+}
+
+impl Hit for axis::Raw {
+    fn hit(&self, input: &Input) -> bool {
+        input.delta.axes_raw.iter().filter(|&&(id, _)| id == self.id).count() > 0
+    }
+}
+
+pub trait HitCount {
+    type Output;
```
alteous: Surprised to see an associated type here.

src/input/mod.rs
vitvakatu: Actually, we can keep it as `Delta`, but then we need to move/rename `Delta` trait. Not a problem, I'm just too lazy to do it and I wanted to check it asap

src/input/mod.rs
vitvakatu: That's because of `axis::Key`

alteous: Oh, never mind... I just saw `type Output = (u8, u8)`.

src/input/mod.rs
```diff
@@ -235,128 +235,148 @@ impl Input {
...
-                Some(-1)
-            } else {
-                None
-            }
-        };
-        let timed_state = state.map(|v| v as f32 * self.delta_time());
-        let mut hits = KeyAxisHits::default();
-        hits.pos = self.delta
+impl HitCount for axis::Key {
+    type Output = (u8, u8);
```
alteous: Nice one. :+1:

src/input/mod.rs
vitvakatu: Not a big deal in my opinion

src/input/mod.rs
```diff
@@ -235,128 +235,148 @@ impl Input {
...
-        let delta_hits = if hits.pos + hits.neg == 0 {
+        (pos, neg)
+    }
+}
+
+pub trait Delta {
+    type Output;
+
+    fn delta(&self, input: &Input) -> Self::Output;
+    fn timed_delta(&self, input: &Input) -> Option<TimerDuration>;
```
alteous: I think `timed` is good enough.

vitvakatu: Yep, maybe. Also we can reduce code a bit by providing default impl for this method

alteous: Are these methods independent? If so, perhaps we should split them into `Delta` and `TimedDelta`.

kvark:
bors r-
wait, is this still WIP?

vitvakatu:
Sorry, I think I need to rebase it. Don't worry, I'll merge with you as reviewer later, can't rebase it now.

kvark:
awesome!
bors r+

Issue #44, opened 2017-06-25
Node children lifetime


kvark:
It appears to be counter-obvious that we have `parent.add(&child)` yet it's not parent holding the child alive but the other way around. In some cases, this goes in the way of encapsulation: e.g. when we load a number of different things from a file, we may want to return a single group that everything is attached to, yet today if you do this - all the children will get removed for losing the references.

There does not appear to be an easy way to address this, on the technical level. The implementation largely prefers storing the parents only and working with them. Ideas:
  1. Keep the children locally by groups and scenes. Problem: hard to synchronize with those children changing parents.
  2. Have `children: Vec<Pointer<Node>>>` alongside `parent: WeakPointer<Node>` in the nodes. Problem: traverse performance, given that accessing children array is expensive, and accessing the parent requires conversion to a strong pointer...
  3. Leave the implementation as is, but change the API to `child.attach(&parent)` to make the relationship clear.

vitvakatu:
I prefer the last solution.

alteous:
I'm not a fan of 2. I didn't quite follow 1, could you explain with a code example?

vitvakatu:
@alteous first one is about storing `Vec<Node>` inside `Group`s, not inside `Hub` (if I understood correctly).

alteous:
@vitvakatu If that's the case then I'm not a fan of 1 neither.

What's quite nice is one can always store the leaf nodes with the parent and use `three_object_wrapper!` to manage them as one entity.

```rust
struct ComplexModel {
    /// Parent of all meshes.
    object: Group,
    /// The meshes (leaf nodes) of the object.
    meshes: Vec<Mesh>,
}
```

alteous:
During #86 I thought about replacing `three_object_wrapper!` with a procedural macro, but later dismissed the idea since it didn't have many tangible benefits. Do you think a `proc_macro` could be used to remedy this issue?

alteous:
The idea was something like
```rust
#[derive(Object)]
struct ComplexModel {
    meshes: Vec<Mesh>,
}
```

vitvakatu:
Pushed high priority because we need to make decision ASAP.

alteous:
>  we need to make decision ASAP.

We do?!

vitvakatu:
@alteous this is serious architectural issue, so we need to solve it to move forward. I'm for `attach` variant, as I said beforehand.

alteous:
[...]

PR #131, opened 2017-10-16
Implement Object::set_parent

alteous:
This PR addresses #44, implementing the `set_parent` idea.


src/render/mod.rs
```diff
@@ -506,12 +506,16 @@ impl Renderer {
     ) {
         self.device.cleanup();
         let mut hub = scene.hub.lock().unwrap();
+        let scene_id;
```
vitvakatu:
```rust
let scene_id = {
    hub.nodes[&scene.object.node].scene_id;
};
```
?

PR #133, opened 2017-10-17
 Make key axis deltas continuous instead of discrete 

alteous:
Fixes unwanted 'jagged' input behaviour. This most notably affects the first person camera controls.

#### Behaviour before fix

Frame | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Event | - | P | - | - | - | P | P | R | - | -
Move | N | Y | N | N | N | Y | Y | N | N | N

#### Behaviour after fix

Frame | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Event | - | P | - | - | - | P | P | R | - | -
Move | N | Y | Y | Y | Y | Y | Y | N | N | N

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -261,14 +261,26 @@ impl Input {
         hit.hit(self)
     }
 
-    /// Returns `Option<delta>` value for specified axis, where `delta` is either `i8` for
-    /// [`axis::Key`](struct.Key.html) or `f32` for [`axis::Raw`](struct.Raw.html).
+    /// Returns the change ('delta') in input state since the last call to
+    /// [`Window::update`].
     ///
-    /// Delta value for [`axis::Key`](struct.Key.html) represents the amount of `positive` hits minus
-    /// amount of `negative` hits.
+    /// This value depends on the type of input device is given.
     ///
-    /// Delta value for [`axis::Raw`](struct.Raw.html) represents the sum of all
-    /// [raw movements](struct.Input.html#method.axes_movements) along specific axis.
+    /// [`axis::Key`]
+    ///
+    ///  * `None` when no updates to the axis are received.
+    ///  * `Some(1)` when only positive input to the axis is received.
+    ///  * `Some(-1)` when only negative input to the axis is received.
+    ///  * `Some(0)` when both positive and negative input to the axis is received.
+    ///
+    /// [`axis::Raw`]
+    ///
+    ///  * `None` when no updates to the axis are received and
+    ///  * `Some(x)` where `x` is the sum of positive and negative inputs otherwise.
+    ///
+    /// [`Window::update`]: window/struct.Window.html#method.update
+    /// [`axis::Key`]: input/axis/struct.Key.html
+    /// [`axis::Raw`]: input/axis/struct.Raw.html
     pub fn delta<D: Delta>(
         &self,
         delta: D,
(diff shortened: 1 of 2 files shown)


src/input/mod.rs
```diff
@@ -329,34 +329,6 @@ impl Hit for Button {
     }
 }
 
-impl Hit for axis::Key {
```
vitvakatu: Maybe we should check for `keys_pressed` here and return `true` in case either `pos` or `neg` button was pressed?

src/input/mod.rs
```diff
@@ -456,8 +428,17 @@ impl Delta for axis::Key {
...
         input: &Input,
     ) -> Option<TimerDuration> {
-        self.delta(input)
-            .map(|v| v as TimerDuration * input.delta_time())
+        let is_pos = input.state.keys_pressed.contains(&self.pos);
+        let is_neg = input.state.keys_pressed.contains(&self.neg);
+        if is_pos && is_neg {
+            Some(0.0)
+        } else if is_pos {
+            Some(input.delta_time())
```
vitvakatu: To make situation clear:
The intent behind `timed` was the shortcut for `timed = delta_time * delta`. And with this approach `timed` and `delta` aren't connected anymore. If it's okay for you from design point of view - I'll accept, just to avoid any misunderstanding.

src/input/mod.rs
alteous: This is where the semantics of `trait Hit` are unclear. Does `hit` mean hit only in the current frame, or otherwise?

src/input/mod.rs
alteous: I don't understand the intention here, sorry! It looks like the time is scaled by the number of hits in one frame. Unless the event rate and/or frame rate is constant I can see this being a problem...

alteous: Games update loops are usually designed to be frame-rate independent.

src/input/mod.rs
vitvakatu: hit fir key means "this key has been pressed and it's in pressed state atm

src/input/mod.rs
vitvakatu: Can you please improve docs in this case?

src/input/mod.rs
alteous: In that case this needs to be amended as suggested. Will fix.

src/input/mod.rs
alteous: Will do.

alteous: I'm really not sure what to do here. Could you advise?

.gitignore
```diff
@@ -5,3 +5,6 @@
 # Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
 # More information here http://doc.crates.io/guide.html#cargotoml-vs-cargolock
 Cargo.lock
+
+.idea/
```
vitvakatu: use yourown global gitignore please, we're not able to track every IDE

src/input/mod.rs
```diff
@@ -333,21 +333,18 @@ impl Hit for axis::Key {
...
+        pos_hit.or(neg_hit).is_some()
     }
 }
 
 impl Hit for axis::Raw {
     fn hit(
         &self,
         input: &Input,
-    ) -> bool {
+    ) > bool {
```
vitvakatu: ?

src/input/mod.rs
vitvakatu: I will take a look a bit later

src/input/mod.rs
alteous: Huh, that's weird. I probably pressed 'x' by mistake before pushing.

vitvakatu:
Hmmm... I think it's all good now. I'm not sure about correctness of `Input::timed` docs. For `axis::Key` this doc string will be valid only for one hit per frame.
So if you want, you can expand doc string for `Input::timed` a bit and cover actual behaviour.
Otherwise just address two previous comments, perform a rustfmt pass and squash.

Thank you for improving my code :D :+1:

alteous:
Just need to finish off the documentation.

PR #135, opened 2017-10-24
Get rid of Node and NodePointer types

vitvakatu:
Also renamed other types:
NodeInfo -> Node
NodeTransform -> Transform


vitvakatu:
Yay, CI passed, ready to review

alteous:
What does @kvark think of these changes?

kvark:
Will have a look in a few hours

PR #136, opened 2017-10-25
Implement multiple traits for all objects

vitvakatu:


src/animation.rs
```diff
@@ -234,6 +236,26 @@ pub struct Action {
...
+}
+
+impl Eq for Action {}
+
+impl Hash for Action {
+    fn hash<H: Hasher>(
+        &self,
+        state: &mut H,
+    ) {
+        self.pointer.hash(state);
```
alteous: Is there any way the value `self.pointer` can change? If not, then all is well.

src/object.rs
```diff
@@ -84,7 +105,7 @@ impl Object {
     }
 
     /// Add new [`Object`](struct.Object.html) to the group.
-    pub fn set_parent<P: Deref<Target=Object>>(
+    pub fn set_parent<P: Deref<Target = Object>>(
         &mut self,
```
alteous: Is `P: AsRef<Object>` achievable instead?

vitvakatu: Checking it now...

src/animation.rs
vitvakatu: I don't think it's possible

src/object.rs
vitvakatu: Replaced, check it out

PR #138, opened 2017-10-26
Use old include rules

### LINKED ISSUE #137: Noisy build messages
Caused by warnings of changing include/exclude rules. See this [Travis log](https://travis-ci.org/three-rs/three/jobs/292852306#L484-L522) for example.

alteous:
Fixes #137

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -1,6 +1,5 @@
 use std::fmt;
 use std::hash::{Hash, Hasher};
-use std::ops::Deref;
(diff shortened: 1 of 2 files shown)


src/object.rs
```diff
@@ -1,6 +1,5 @@
 use std::fmt;
 use std::hash::{Hash, Hasher};
-use std::ops::Deref;
```
alteous: Was unused.

PR #144, opened 2017-10-29
Implement SyncGuard

### LINKED ISSUE #134: Massive `sync`s
It's common situation while programming game when you need to update information about multiple objects. For example, update positions of every enemy and cache in `Enemy` type for future using.

But the `Object::sync` is too slow because it updates the whole scene graph every time it's called.
So I present rough idea how to handle these situations:
```rust
// `Enemy` must implement `AsRef<Object>`
let mut enemies: Vec<Enemy> = create_enemies();
//...
// somewhere in main loop
// `with_sync` will update scene graph once
let sync = window.scene.with_sync();
// First argument is iterator, second one is working closure.
sync.update(&mut enemies, |_, node_info| { println!("Enemy at: {}", node_info.world_transform.position); });
sync.update(&mut enemies, |mut enemy, node_info| { enemy.position = node_info.world_transform.position; });
```

There are a lot of questions about implementation (for example, how to allow to pass any iterator in), but the core idea should be clear.

What do you think?

**UPD: code example updated after short discussion on gitter**

vitvakatu:
Closes #134

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -12,7 +12,7 @@ pub(crate) type TransformInternal = cgmath::Decomposed<cgmath::Vector3<f32>, cgm
 
 // Fat node of the scene graph.
 //
-// `Node` is used by `three-rs` internally,
+// `NodeInternal` is used by `three-rs` internally,
 // client code uses [`Object`](struct.Object.html) instead.
 #[derive(Debug)]
 pub(crate) struct NodeInternal {
@@ -30,6 +30,21 @@ pub(crate) struct NodeInternal {
     pub(crate) scene_id: Option<scene::Uid>,
     /// Context specific-data, for example, `UiText`, `Visual` or `Light`.
     pub(crate) sub_node: SubNode,
+}
+
+impl NodeInternal {
+    pub(crate) fn to_node(&self) -> Node {
+        Node {
+            transform: self.transform.into(),
+            world_transform: self.world_transform.into(),
+            visible: self.visible,
+            world_visible: self.world_visible,
+            material: match self.sub_node {
+                SubNode::Visual(ref mat, _) => Some(mat.clone()),
+                _ => None,
+            },
+        }
+    }
 }
 
 /// Position, rotation, and scale of the scene `Node`.
--- a/src/scene.rs
+++ b/src/scene.rs
@@ -1,8 +1,11 @@
-
+//! `Scene` and `SyncGuard` structures.
 use color::Color;
-use hub::HubPtr;
+use hub::{Hub, HubPtr};
+use node::Node;
 use object::Object;
 use texture::{CubeMap, Texture};
+
+use std::sync::MutexGuard;
 
 /// Unique identifier for a scene.
 pub type Uid = usize;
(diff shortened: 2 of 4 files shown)


src/scene.rs
```diff
@@ -28,4 +29,115 @@ pub struct Scene {
...
+            closure(item, node);
+        })
+    }
+}
+
+impl Scene {
+    /// Create new [`SyncGuard`](struct.SyncGuard.html).
+    ///
+    /// This is performance-costly operation, you should not use it many times per frame.
+    pub fn with_sync<'a>(&'a mut self) -> SyncGuard<'a> {
```
alteous: `with_sync` is an odd-sounding name. May I suggest `sync_guard` instead?

src/scene.rs
```diff
@@ -28,4 +29,115 @@ pub struct Scene {
...
+        let scene_id = {
+            let mut hub = self.hub.lock().unwrap();
```
alteous: Doesn't matter too much at the moment but it might be worth keeping hold of the lock for the lifetime of `SyncGuard`.

src/scene.rs
```diff
@@ -28,4 +29,115 @@ pub struct Scene {
...
+impl<'a> SyncGuard<'a> {
+    /// Walk through every item in `iterator`, obtain its [`Node`] and call user-supplied closure.
+    ///
+    /// # Panics
+    /// Panics if `scene` doesn't have this `Object`.
+    ///
+    /// [`Node`]: ../node/struct.Node.html
+    pub fn update<T: 'a, I, F>(
+        &mut self,
+        iterator: I,
```
alteous: In future, it might be beneficial to have an `Extend` implementation, so the user can pass the guard around and iterate over different types.

src/scene.rs
vitvakatu: Naming is always the problem...
Yep, I thought about something else, I think I'll change to `sync_guard`

src/scene.rs
vitvakatu: Mm... I'm not quite following how `Extend` can be be useful

alteous: The idea I had in mind was to make `SyncGuard` like a collection of objects to be synchronised. Perhaps `Extend` isn't a great way to do this. This is what I had in mind:

```rust
let mut sharks = vec![Shark];
let mut lasers = vec![Laser];
{
    scene.sync_guard()
        .update(&mut sharks, |shark, node| shark.attach_to_head(Laser))
        .update(&mut laser, |laser, node| laser.fire(node.transform))
        .exec();
}
```

src/scene.rs
alteous: > Naming is always the problem...

Sorry to be pedantic/bikesheddy. :innocent:

vitvakatu:
Finally got some time to finish this stuff.

src/scene.rs
vitvakatu: I don't think it's better than current design proposal. You can do the same thing with multiple `update`s, and without any performance issues.

kvark:
Please rebase

vitvakatu:
Yay, rebased!

vitvakatu:
@kvark any concerns? I think three weeks is a great time for PR.

src/node.rs
```diff
@@ -69,6 +69,21 @@ pub struct Node {
     pub material: Option<Material>,
 }
 
+impl<'a> From<&'a NodeInternal> for Node {
```
kvark: any reason this is not just a `to_node(&self)` method in `NodeInternal`?

src/scene.rs
```diff
@@ -28,4 +31,109 @@ pub struct Scene {
...
+        mut closure: F,
+    ) where
+        T: AsRef<Object>,
+        I: Iterator<Item = &'a mut T>,
+        F: FnMut(&'a mut T, Node),
+    {
+        iterator.for_each(|item| {
```
kvark: essentially the logic of this method comes down to providing `fn(&Object) -> Node`. Having this (for example, as `SyncGuard::resolve()` method) the user can just do the iteration themselves:
```rust
for item in iterator {
   let node = sync_guard.resolve(item);
   closure(item, node);
}
```

src/node.rs
vitvakatu: No particular reason except for the common way of expressing translation from one type to another using `From` trait. Any reason not to do that?

src/scene.rs
vitvakatu: I think you're right, sounds reasonable.

alteous:
Sorry for the delay. Looks good to me.

kvark:
Creating another object from a reference of self using From is not idiomatic.

kvark:
Actually, instead of having this function in the SyncGuard struct, it should just straight implement the trait for this function.

vitvakatu:
@kvark I'm not quite following your latter idea. 
> Actually, instead of having this function in the SyncGuard struct, it should just straight implement the trait for this function.

kvark:
Instead of having "impl SyncGuard { fn foo() {..}" do "impl Fn(...) for SyncGuard {..}"

vitvakatu:
@kvark your objections were addressed. Implementing `Fn` or `FnMut` for `SyncGuard` is impossible in current stable, so we still need to use `resolve` function.
**Do not merge, needs squashing**

vitvakatu:
bors r=kvark

PR #150, opened 2017-12-09
Auto instancing

### LINKED ISSUE #37: Auto-instancing
We have just a handful of different PSOs, and we can have lots of draw calls. Given that we populate all the local data anyway, we could build it with instanced buffers and issue fewer calls.
Related - #36

vitvakatu:
fixes #37 

Q&A:
## Insanity? 
No, just auto instancing support. *(I'm lying, it's crazy)*
## What have you done?
Short list:
- All shaders were modified - no more `Locals`, just vertex attributes.
- ~~`Material` enum was renamed to `MaterialType`, made private, added new `Material` struct.
Why? Because I need UIDs to group meshes in renderer.~~
- Huge refactoring in the `render` module in order to reduce code duplication for both instanced and regular rendering.
- ~~`mesh_instance*` was renamed to `duplicate_mesh*`. Reason - you can't change material for mesh with instancing support. (Well, actually you can, but you will lose optimization. Perhaps I should keep this method?)~~

## Worth it?
Totally! In group example, performance have increased from 14 to stable 60 FPS (in release mode).

Some things looks really weird, so ask me please.
Renderer now looks like a total mess, but I would like to continue my work on it.

### DIFF
--- a/data/shaders/basic_vs.glsl
+++ b/data/shaders/basic_vs.glsl
@@ -1,13 +1,22 @@
 #version 150 core
-#include <locals>
 #include <globals>
 
 in vec4 a_Position;
 in vec4 a_Normal;
 in vec2 a_TexCoord;
 out vec2 v_TexCoord;
+out vec4 v_Color;
+
+in vec4 i_World0;
+in vec4 i_World1;
+in vec4 i_World2;
+in vec4 i_World3;
+in vec4 i_Color;
+in vec4 i_UvRange;
 
 void main() {
-    v_TexCoord = mix(u_UvRange.xy, u_UvRange.zw, a_TexCoord);
-    gl_Position = u_ViewProj * u_World * a_Position;
+    mat4 m_World = mat4(i_World0, i_World1, i_World2, i_World3);
+    v_TexCoord = mix(i_UvRange.xy, i_UvRange.zw, a_TexCoord);
+    v_Color = i_Color;
+    gl_Position = u_ViewProj * m_World * a_Position;
 }
--- a/data/shaders/basic_vs.glsl
+++ b/data/shaders/basic_vs.glsl
@@ -10,12 +10,11 @@ out vec4 v_Color;
 in vec4 i_World0;
 in vec4 i_World1;
 in vec4 i_World2;
-in vec4 i_World3;
 in vec4 i_Color;
 in vec4 i_UvRange;
 
 void main() {
-    mat4 m_World = mat4(i_World0, i_World1, i_World2, i_World3);
+    mat4 m_World = transpose(mat4(i_World0, i_World1, i_World2, vec4(0.0, 0.0, 0.0, 1.0)));
     v_TexCoord = mix(i_UvRange.xy, i_UvRange.zw, a_TexCoord);
     v_Color = i_Color;
     gl_Position = u_ViewProj * m_World * a_Position;
--- a/examples/group.rs
+++ b/examples/group.rs
@@ -132,7 +132,7 @@ fn main() {
     let mut fps_counter = win.factory.ui_text(&font, "FPS: 00");
 
     let timer = win.input.time();
-    println!("{}", cubes.len());
+    println!("Total number of cubes: {}", cubes.len());
     while win.update() && !win.input.hit(three::KEY_ESCAPE) {
         let time = timer.get(&win.input);
         let delta_time = win.input.delta_time();
--- a/examples/group.rs
+++ b/examples/group.rs
@@ -132,9 +132,14 @@ fn main() {
     let mut cubes = create_cubes(&mut win.factory, &materials, &levels);
     cubes[0].group.set_parent(&win.scene);
 
+    let font = win.factory.load_font(format!("{}/data/fonts/DejaVuSans.ttf", env!("CARGO_MANIFEST_DIR")));
+    let mut fps_counter = win.factory.ui_text(&font, "FPS: 00");
+
     let timer = win.input.time();
     while win.update() && !win.input.hit(three::KEY_ESCAPE) {
         let time = timer.get(&win.input);
+        let delta_time = win.input.delta_time();
+        fps_counter.set_text(format!("FPS: {}", 1.0 / delta_time));
         for cube in cubes.iter_mut() {
             let level = &levels[cube.level_id];
             let angle = Rad(time * level.speed);
--- a/examples/group.rs
+++ b/examples/group.rs
@@ -104,14 +104,7 @@ fn create_cubes(
 
 const COLORS: [three::Color; 6] = [0xffff80, 0x8080ff, 0x80ff80, 0xff8080, 0x80ffff, 0xff80ff];
 
-const SPEEDS: [f32; 5] = [
-    0.7,
-    -1.0,
-    1.3,
-    -1.6,
-    1.9,
-    //-2.2, //TODO when performance allows
-];
+const SPEEDS: [f32; 6] = [0.7, -1.0, 1.3, -1.6, 1.9, -2.2];
 
 fn main() {
     let mut win = three::Window::new("Three-rs group example");
@@ -132,10 +125,14 @@ fn main() {
     let mut cubes = create_cubes(&mut win.factory, &materials, &levels);
     cubes[0].group.set_parent(&win.scene);
 
-    let font = win.factory.load_font(format!("{}/data/fonts/DejaVuSans.ttf", env!("CARGO_MANIFEST_DIR")));
+    let font = win.factory.load_font(format!(
+        "{}/data/fonts/DejaVuSans.ttf",
+        env!("CARGO_MANIFEST_DIR")
+    ));
     let mut fps_counter = win.factory.ui_text(&font, "FPS: 00");
 
     let timer = win.input.time();
+    println!("{}", cubes.len());
     while win.update() && !win.input.hit(three::KEY_ESCAPE) {
         let time = timer.get(&win.input);
         let delta_time = win.input.delta_time();
--- a/src/factory/mod.rs
+++ b/src/factory/mod.rs
@@ -119,7 +119,6 @@ impl Factory {
             .unwrap()
     }
 
-
     pub(crate) fn new(mut backend: BackendFactory) -> Self {
         let quad_buf = backend.create_vertex_buffer(&QUAD);
         let default_sampler = backend.create_sampler_linear();
@@ -230,13 +229,11 @@ impl Factory {
             )
         };
         izip!(position_iter, normal_iter, tangent_iter, uv_iter)
-            .map(|(position, normal, tangent, tex_coord)| {
-                Vertex {
-                    pos: [position.x, position.y, position.z, 1.0],
-                    normal: normal,
-                    uv: tex_coord,
-                    tangent: tangent,
-                }
+            .map(|(position, normal, tangent, tex_coord)| Vertex {
+                pos: [position.x, position.y, position.z, 1.0],
+                normal: normal,
+                uv: tex_coord,
+                tangent: tangent,
             })
             .collect()
     }
@@ -342,8 +339,8 @@ impl Factory {
             SubNode::Visual(_, ref gpu) => GpuData {
                 instances,
                 instance_cache_key: Some(InstanceCacheKey {
-                    material_id: material.id,
-                    geometry_id: gpu.vertices.clone(),
+                    material: material.clone(),
+                    geometry: gpu.vertices.clone(),
                 }),
                 ..gpu.clone()
             },
@@ -368,8 +365,8 @@ impl Factory {
             SubNode::Visual(_, ref gpu) => GpuData {
                 instances,
                 instance_cache_key: Some(InstanceCacheKey {
-                    material_id: material.id,
-                    geometry_id: gpu.vertices.clone(),
+                    material: material.clone(),
+                    geometry: gpu.vertices.clone(),
                 }),
                 ..gpu.clone()
             },
@@ -677,23 +674,14 @@ impl Factory {
             .iter()
             .map(|path| {
                 let format = Factory::parse_texture_format(path.as_ref());
-                let file = fs::File::open(path).unwrap_or_else(|e| {
-                    panic!("Unable to open {}: {:?}", path.as_ref().display(), e)
-                });
+                let file = fs::File::open(path).unwrap_or_else(|e| panic!("Unable to open {}: {:?}", path.as_ref().display(), e));
                 image::load(io::BufReader::new(file), format)
-                    .unwrap_or_else(|e| {
-                        panic!("Unable to decode {}: {:?}", path.as_ref().display(), e)
-                    })
+                    .unwrap_or_else(|e| panic!("Unable to decode {}: {:?}", path.as_ref().display(), e))
                     .to_rgba()
             })
             .collect::<Vec<_>>();
         let data: [&[u8]; 6] = [
-            &images[0],
-            &images[1],
-            &images[2],
-            &images[3],
-            &images[4],
-            &images[5],
+            &images[0], &images[1], &images[2], &images[3], &images[4], &images[5]
         ];
         let size = images[0].dimensions().0;
         let kind = t::Kind::Cube(size as t::Size);
@@ -871,9 +859,7 @@ impl Factory {
 
                 info!(
                     "\tmaterial {} with {} normals and {} uvs",
-                    gr.name,
-                    num_normals,
-                    num_uvs
+                    gr.name, num_normals, num_uvs
                 );
                 let material = match gr.material {
                     Some(ref rc_mat) => self.load_obj_material(&*rc_mat, num_normals != 0, num_uvs != 0, path_parent),
--- a/src/factory/mod.rs
+++ b/src/factory/mod.rs
@@ -31,7 +31,7 @@ use light::{Ambient, Directional, Hemisphere, Point, ShadowMap};
 use material::Material;
 use mesh::{DynamicMesh, Mesh};
 use object::{Group, Object};
-use render::{basic_pipe, BackendFactory, BackendResources, BasicPipelineState, DynamicData, GpuData, InstanceCacheKey, ShadowFormat, Vertex};
+use render::{basic_pipe, BackendFactory, BackendResources, BasicPipelineState, DynamicData, GpuData, Instance, InstanceCacheKey, ShadowFormat, Vertex};
 use scene::Scene;
 use sprite::Sprite;
 use text::{Font, Text, TextData};
@@ -107,6 +107,19 @@ fn f2i(x: f32) -> I8Norm {
 }
 
 impl Factory {
+    fn create_instance_buffer(&mut self) -> gfx::handle::Buffer<BackendResources, Instance> {
+        // TODO: Better error handling
+        self.backend
+            .create_buffer(
+                1,
+                gfx::buffer::Role::Vertex,
+                gfx::memory::Usage::Dynamic,
+                gfx::TRANSFER_DST,
+            )
+            .unwrap()
+    }
+
+
     pub(crate) fn new(mut backend: BackendFactory) -> Self {
         let quad_buf = backend.create_vertex_buffer(&QUAD);
         let default_sampler = backend.create_sampler_linear();
@@ -235,14 +248,7 @@ impl Factory {
         material: M,
     ) -> Mesh {
         let vertices = Self::mesh_vertices(&geometry.base_shape);
-        let instances = self.backend
-            .create_buffer(
-                1,
-                gfx::buffer::Role::Vertex,
-                gfx::memory::Usage::Dynamic,
-                gfx::TRANSFER_DST,
-            )
-            .unwrap();
+        let instances = self.create_instance_buffer();
         let (vbuf, mut slice) = if geometry.faces.is_empty() {
             self.backend.create_vertex_buffer_with_slice(&vertices, ())
         } else {
@@ -252,7 +258,6 @@ impl Factory {
         };
         slice.instances = Some((1, 0));
         let material = material.into();
-        let material_id = material.id;
         Mesh {
             object: self.hub.lock().unwrap().spawn_visual(
                 material,
@@ -261,11 +266,7 @@ impl Factory {
                     vertices: vbuf.clone(),
                     instances,
                     pending: None,
-                    use_instancing: false,
-                    instance_cache_key: InstanceCacheKey {
-                        geometry_id: vbuf.clone(),
-                        material_id,
-                    },
+                    instance_cache_key: None,
                 },
             ),
         }
@@ -303,16 +304,8 @@ impl Factory {
             }
             (data.len(), dest_buf, upload_buf)
         };
-        let instances = self.backend
-            .create_buffer(
-                1,
-                gfx::buffer::Role::Vertex,
-                gfx::memory::Usage::Dynamic,
-                gfx::TRANSFER_DST,
-            )
-            .unwrap();
+        let instances = self.create_instance_buffer();
         let material = material.into();
-        let material_id = material.id;
         DynamicMesh {
             object: self.hub.lock().unwrap().spawn_visual(
                 material,
@@ -321,11 +314,7 @@ impl Factory {
                     vertices: vertices.clone(),
                     instances,
                     pending: None,
-                    use_instancing: false,
-                    instance_cache_key: InstanceCacheKey {
-                        geometry_id: vertices.clone(),
-                        material_id,
-                    },
+                    instance_cache_key: None,
                 },
             ),
             geometry,
@@ -343,25 +332,21 @@ impl Factory {
         &mut self,
         template: &Mesh,
     ) -> Mesh {
+        let instances = self.create_instance_buffer();
         let mut hub = self.hub.lock().unwrap();
-        let instances = self.backend
-            .create_buffer(
-                1,
-                gfx::buffer::Role::Vertex,
-                gfx::memory::Usage::Dynamic,
-                gfx::TRANSFER_DST,
-            )
-            .unwrap();
+        let material = match hub.get(&template).sub_node {
+            SubNode::Visual(ref mat, _) => mat.clone(),
+            _ => unreachable!(),
+        };
         let gpu_data = match hub.get(&template).sub_node {
             SubNode::Visual(_, ref gpu) => GpuData {
                 instances,
-                use_instancing: true,
+                instance_cache_key: Some(InstanceCacheKey {
+                    material_id: material.id,
+                    geometry_id: gpu.vertices.clone(),
+                }),
                 ..gpu.clone()
             },
-            _ => unreachable!(),
-        };
-        let material = match hub.get(&template).sub_node {
-            SubNode::Visual(ref mat, _) => mat.clone(),
             _ => unreachable!(),
         };
         Mesh {
@@ -376,25 +361,22 @@ impl Factory {
         template: &Mesh,
         material: M,
     ) -> Mesh {
+        let instances = self.create_instance_buffer();
         let mut hub = self.hub.lock().unwrap();
-        let instances = self.backend
-            .create_buffer(
-                1,
-                gfx::buffer::Role::Vertex,
-                gfx::memory::Usage::Dynamic,
-                gfx::TRANSFER_DST,
-            )
-            .unwrap();
+        let material = material.into();
         let gpu_data = match hub.get(&template).sub_node {
             SubNode::Visual(_, ref gpu) => GpuData {
                 instances,
-                use_instancing: true,
+                instance_cache_key: Some(InstanceCacheKey {
+                    material_id: material.id,
+                    geometry_id: gpu.vertices.clone(),
+                }),
                 ..gpu.clone()
             },
             _ => unreachable!(),
         };
         Mesh {
-            object: hub.spawn_visual(material.into(), gpu_data),
+            object: hub.spawn_visual(material, gpu_data),
         }
     }
 
@@ -403,18 +385,10 @@ impl Factory {
         &mut self,
         material: material::Sprite,
     ) -> Sprite {
-        let instances = self.backend
-            .create_buffer(
-                1,
-                gfx::buffer::Role::Vertex,
-                gfx::memory::Usage::Dynamic,
-                gfx::TRANSFER_DST,
-            )
-            .unwrap();
+        let instances = self.create_instance_buffer();
         let mut slice = gfx::Slice::new_match_vertex_buffer(&self.quad_buf);
         slice.instances = Some((1, 0));
         let material = Material::from(material);
-        let material_id = material.id;
         Sprite::new(self.hub.lock().unwrap().spawn_visual(
             material,
             GpuData {
@@ -422,11 +396,7 @@ impl Factory {
                 vertices: self.quad_buf.clone(),
                 instances,
                 pending: None,
-                use_instancing: false,
-                instance_cache_key: InstanceCacheKey {
-                    geometry_id: self.quad_buf.clone(),
-                    material_id,
-                },
+                instance_cache_key: None,
             },
         ))
     }
@@ -913,7 +883,6 @@ impl Factory {
                     }.into(),
                 };
                 info!("\t{:?}", material);
-                let material_id = material.id;
 
                 let (vbuf, mut slice) = self.backend
                     .create_vertex_buffer_with_slice(&vertices, &indices[..]);
@@ -934,11 +903,7 @@ impl Factory {
                             vertices: vbuf.clone(),
                             instances,
                             pending: None,
-                            use_instancing: false,
-                            instance_cache_key: InstanceCacheKey {
-                                geometry_id: vbuf.clone(),
-                                material_id,
-                            },
+                            instance_cache_key: None,
                         },
                     ),
                 };
--- a/src/factory/mod.rs
+++ b/src/factory/mod.rs
@@ -246,7 +246,7 @@ impl Factory {
     ) -> Mesh {
         let vertices = Self::mesh_vertices(&geometry.base_shape);
         let instances = self.create_instance_buffer();
-        let (vbuf, mut slice) = if geometry.faces.is_empty() {
+        let (vertices, mut slice) = if geometry.faces.is_empty() {
             self.backend.create_vertex_buffer_with_slice(&vertices, ())
         } else {
             let faces: &[u32] = gfx::memory::cast_slice(&geometry.faces);
@@ -254,13 +254,12 @@ impl Factory {
                 .create_vertex_buffer_with_slice(&vertices, faces)
         };
         slice.instances = Some((1, 0));
-        let material = material.into();
         Mesh {
             object: self.hub.lock().unwrap().spawn_visual(
-                material,
+                material.into(),
                 GpuData {
                     slice,
-                    vertices: vbuf.clone(),
+                    vertices,
                     instances,
                     pending: None,
                     instance_cache_key: None,
@@ -302,13 +301,12 @@ impl Factory {
             (data.len(), dest_buf, upload_buf)
         };
         let instances = self.create_instance_buffer();
-        let material = material.into();
         DynamicMesh {
             object: self.hub.lock().unwrap().spawn_visual(
-                material,
+                material.into(),
                 GpuData {
                     slice,
-                    vertices: vertices.clone(),
+                    vertices,
                     instances,
                     pending: None,
                     instance_cache_key: None,
@@ -870,7 +868,7 @@ impl Factory {
                 };
                 info!("\t{:?}", material);
 
-                let (vbuf, mut slice) = self.backend
+                let (vertices, mut slice) = self.backend
                     .create_vertex_buffer_with_slice(&vertices, &indices[..]);
                 slice.instances = Some((1, 0));
                 let instances = self.backend
@@ -886,7 +884,7 @@ impl Factory {
                         material,
                         GpuData {
                             slice,
-                            vertices: vbuf.clone(),
+                            vertices,
                             instances,
                             pending: None,
                             instance_cache_key: None,
--- a/src/factory/mod.rs
+++ b/src/factory/mod.rs
@@ -31,7 +31,7 @@ use light::{Ambient, Directional, Hemisphere, Point, ShadowMap};
 use material::Material;
 use mesh::{DynamicMesh, Mesh};
 use object::{Group, Object};
-use render::{basic_pipe, BackendFactory, BackendResources, BasicPipelineState, DynamicData, GpuData, ShadowFormat, Vertex};
+use render::{basic_pipe, BackendFactory, BackendResources, BasicPipelineState, DynamicData, GpuData, InstanceCacheKey, ShadowFormat, Vertex};
 use scene::Scene;
 use sprite::Sprite;
 use text::{Font, Text, TextData};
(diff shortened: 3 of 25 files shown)


alteous:
Holy crap this is impressive! Totally worth the wait. :)

kvark:
> Reason - you can't change material for mesh with instancing support. (Well, actually you can, but you will lose optimization. Perhaps I should keep this method?)

I'd prefer the original method. Let's define the optimizations to be at the implementation discretion. I.e. user should still be able to change the material properties (as opposed to changing to a different material entirely) without the loss of optimization.

data/shaders/basic_vs.glsl
```diff
@@ -1,13 +1,22 @@
...
-#include <locals>
 #include <globals>
 
 in vec4 a_Position;
 in vec4 a_Normal;
 in vec2 a_TexCoord;
 out vec2 v_TexCoord;
+out vec4 v_Color;
+
+in vec4 a_World0;
```
kvark: how about adopting a different prefix for the instance variables? say, `i_XXX`.

examples/group.rs
```diff
@@ -104,14 +108,7 @@ fn create_cubes(
...
 
-const SPEEDS: [f32; 5] = [
-    0.7,
-    -1.0,
-    1.3,
-    -1.6,
-    1.9,
-    //-2.2, //TODO when performance allows
-];
+const SPEEDS: [f32; 6] = [0.7, -1.0, 1.3, -1.6, 1.9, -2.2];
```
kvark: 🎉

src/factory/mod.rs
```diff
@@ -334,15 +370,23 @@ impl Factory {
...
     /// Create a `Mesh` sharing the geometry with another one but with a different material.
     /// Rendering a sequence of meshes with the same geometry is faster.
-    pub fn mesh_instance_with_material<M: Into<Material>>(
+    pub fn duplicate_mesh_with_material<M: Into<Material>>(
         &mut self,
         template: &Mesh,
         material: M,
     ) -> Mesh {
         let mut hub = self.hub.lock().unwrap();
+        let instances = self.backend
```
kvark: might want to have a helper private method in the factory for this, given how many times it's repeated

src/geometry.rs
```diff
@@ -70,6 +73,7 @@ pub struct Geometry {
     pub shapes: HashMap<String, Shape>,
     /// Faces.
     pub faces: Vec<[u32; 3]>,
+    #[doc(hidden)] pub __id: usize,
```
kvark: does it have to be public?

src/mesh.rs
```diff
@@ -107,11 +107,23 @@ impl Hash for DynamicMesh {
...
-        let msg = Operation::SetMaterial(material);
+        let msg = Operation::SetMaterial(material.into());
+        let _ = self.object.tx.send((self.object.node.downgrade(), msg));
+    }
+
+    /// Enable or disable instanced rendering for this `Mesh`.
+    ///
+    /// It will improve performance in case of multiple `Mesh`es with the same
+    /// materials and geometries.
+    pub fn use_instancing(
```
kvark: I'd like us to not have this explicit switch, preferably. We can technically draw everything with instancing. Whatever happens to never be instanced will just have an instance of 1. Whatever does get instanced with `instance_mesh` should be automatically fast.

src/render/mod.rs
```diff
@@ -434,19 +466,29 @@ impl Renderer {
         let quad_buf = gl_factory.create_constant_buffer(1);
         let light_buf = gl_factory.create_constant_buffer(MAX_LIGHTS);
         let pbr_buf = gl_factory.create_constant_buffer(1);
+        let inst_buf = gl_factory
+            .create_buffer(
+                INSTANCE_COUNT,
```
kvark: instead of keeping around a fixed-inst-count buffer, I think we should have an automatically grown instance buffer per shader. We collect CPU instances first, then see if the buffer needs to be re-allocated (and do this), then fill out instance data.

src/render/mod.rs
```diff
@@ -953,6 +939,109 @@ impl Renderer {
         self.encoder.flush(&mut self.device);
     }
 
+    #[inline]
+    fn instance_basic(
```
kvark: maybe have those constructors inside `impl Instance`?

vitvakatu: yes, it'd be better

data/shaders/basic_vs.glsl
vitvakatu: I'm not familiar with naming conventions in shaders so yes, will do

src/render/mod.rs
vitvakatu: If so, we need to store factory inside Renderer. Can't remember whether it's possible (probably yes)

src/geometry.rs
vitvakatu: Unfortunately :( You can't use nice `.. Default::default() ` syntax on structs with private fields

vitvakatu:
About IDs: you need to group meshes with same geometry and material in renderer. But both geometry and material structs are unhashable (and you can't do them hashable without some unique field). Maybe there's another solution?

vitvakatu:
Okay, all concerns from @kvark were covered. Awaiting further review

data/shaders/basic_vs.glsl
```diff
@@ -1,13 +1,22 @@
...
-#include <locals>
 #include <globals>
 
 in vec4 a_Position;
 in vec4 a_Normal;
 in vec2 a_TexCoord;
 out vec2 v_TexCoord;
+out vec4 v_Color;
+
+in vec4 i_World0;
```
kvark: would it make sense to move those instance inputs into a header ("locals")? hint: it will not if different shaders have different instance inputs

data/shaders/basic_vs.glsl
```diff
@@ -1,13 +1,22 @@
...
+in vec4 i_World1;
+in vec4 i_World2;
+in vec4 i_World3;
```
kvark: as I mentioned before (and I don't think there was any reply?), we should use 3 vectors instead of 4 for the world transformation.

src/material.rs
```diff
@@ -225,80 +248,46 @@ pub struct Wireframe {
...
-
-impl From<Phong> for Material {
-    fn from(params: Phong) -> Material {
-        Material::Phong(params)
-    }
+/// Specifies the appearance of a [`Mesh`](struct.Mesh.html).
+#[derive(Debug, Clone, PartialEq)]
+pub struct Material {
+    pub(crate) mat_type: MaterialType,
+    pub(crate) id: usize,
```
kvark: can we just hash the material instead of introducing the ID?

src/render/mod.rs
```diff
@@ -193,19 +202,71 @@ gfx_defines! {
...
+}
+
+impl Instance {
+    #[inline]
+    fn basic(
+        mx_world: [[f32; 4]; 4],
+        color: u32,
+        uv_range: [f32; 4],
+        param: f32,
+    ) -> Instance {
```
kvark: nit: `Self`

src/render/mod.rs
```diff
@@ -193,19 +202,71 @@ gfx_defines! {
...
 #[derive(Clone, Debug)]
 pub(crate) struct GpuData {
     pub slice: gfx::Slice<back::Resources>,
-    pub vertices: gfx::handle::Buffer<back::Resources, Vertex>,
-    pub constants: gfx::handle::Buffer<back::Resources, Locals>,
+    pub vertices: h::Buffer<back::Resources, Vertex>,
+    pub instances: h::Buffer<back::Resources, Instance>,
     pub pending: Option<DynamicData>,
+    pub use_instancing: bool,
+    pub instance_cache_key: InstanceCacheKey,
```
kvark: could we have `Option<InstanceCacheKey>` for clearer semantics than a separate boolean flag?

src/render/mod.rs
```diff
@@ -953,6 +987,76 @@ impl Renderer {
...
+        mat_type: &MaterialType,
+        shadow_sampler: &h::Sampler<back::Resources>,
+        shadow0: &h::ShaderResourceView<back::Resources, f32>,
+        shadow1: &h::ShaderResourceView<back::Resources, f32>,
+    ) {
+        encoder.update_buffer(&inst_buf, instances, 0).unwrap();
+
+        let slice = if instances.len() > 1 {
+            let mut slice = slice;
+            slice.instances = Some((instances.len() as u32, 0));
```
kvark: oh my, so many `slice` things!
could do `Slice { instances: <something>, ..slice }`

src/render/pso_data.rs
```diff
@@ -0,0 +1,163 @@
...
+    normal: Option<Texture<[f32; 4]>>,
+    emissive: Option<Texture<[f32; 4]>>,
+    metallic_roughness: Option<Texture<[f32; 4]>>,
+    occlusion: Option<Texture<[f32; 4]>>,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct PbrMapParams {
+    pub(crate) base_color: (
+        h::ShaderResourceView<BackendResources, [f32; 4]>,
```
kvark: let's have a type for shader view + sampler, given that we need it many times here

data/shaders/basic_vs.glsl
vitvakatu: Well, most of the shaders use it, so I'll move it in header.

data/shaders/basic_vs.glsl
vitvakatu: Will do

src/material.rs
vitvakatu: No, we can't. Materials contains `f32`s, that are unhashable. Even if we can find a way around it, our `Texture<[f32; 4]>` is unhashable too because of `f32` again.

kvark: `f32` in general are unhashable, but only because of the exceptions like de-normalized floats and NaN, Inf, etc. We can safely assume the f32 in the material are not exceptional (and abort otherwise). Hashing can't be derived automatically, but it can be implemented using one of the following methods:
  - just hashing raw 32bits of the float
  - if the float is known to be in a rage (say, 0-1), quantitize it with enough precision (`u32` is more than enough). 

`Texture<[f32; 4]>` is just a handle, once again, that is hashable. Caveat: we need to investigate in depth the safety of hashing the gfx handles, but for now this is an acceptable choice.

src/material.rs
```diff
@@ -54,6 +62,8 @@ pub mod basic {
 }
 
 /// Parameters for a Lamberian diffusion reflection model.
+///
+/// Renders triangle meshes with the Gouraud illumination model.
```
alteous: *Lambertian illumination model.

Gouraud _shading_ refers to whether lighting computed at the _vertices_ of a polygon are interpolated across the face, hence the `flat` field disables/enables Gouraud shading.

alteous: I considered having an `enum Shading { Flat, Gouraud }` but that felt a bit overkill.

src/render/mod.rs
```diff
@@ -378,6 +425,21 @@ impl PipelineStates {
...
+        material: &MaterialType,
+    ) -> &BasicPipelineState {
+        match *material {
+            MaterialType::Basic(_) => &self.mesh_basic_fill,
+            MaterialType::Line(_) => &self.line_basic,
+            MaterialType::Wireframe(_) => &self.mesh_basic_wireframe,
+            MaterialType::Lambert(_) => &self.mesh_gouraud,
+            MaterialType::Phong(_) => &self.mesh_phong,
+            MaterialType::Sprite(_) => &self.sprite,
+            _ => unreachable!(),
```
alteous: Is it really unreachable or just an internal error if we hit this branch?

data/shaders/basic_vs.glsl
alteous: The instance variables are likely to change, so I'd say not.

src/material.rs
vitvakatu: Sure, I've been moving doc lines forth and back, so it's not the last change.

src/render/mod.rs
vitvakatu: Unreachable for sure (if somebody won't break the renderer logic).

vitvakatu:
Yep, a few more things - safe wrappers around floats and grouping world matrix in 3 vectors.

src/safe_float.rs
```diff
@@ -0,0 +1,38 @@
...
+		Self::try_from(value).expect("Illegal float value")
+	}
+}
+
+impl Eq for SafeFloat {}
+
+impl Hash for SafeFloat {
+	fn hash<H: Hasher>(&self, state: &mut H) {
+        let mut buf = [0; 4];
+        LittleEndian::write_f32(&mut buf, self.0);
```
alteous: Lol

kvark:
Still WIP?

vitvakatu:
Still WIP :(
Needs one last thing - correct Hash implementation for Custom material.

src/render/mod.rs
```diff
@@ -786,6 +786,7 @@ impl Renderer {
             };
 
             let mx_world: [[f32; 4]; 4] = Matrix4::from(node.world_transform).into();
+            println!("{:?}", mx_world);
```
kvark: debug leftover

src/safe_float.rs
```diff
@@ -0,0 +1,38 @@
+use std::num::FpCategory;
+use std::hash::{Hash, Hasher};
+use std::ops::Deref;
+use byteorder::{ByteOrder, LittleEndian};
+
+#[derive(Debug, Clone, Copy, Default, PartialEq)]
+pub struct SafeFloat(f32);
+
+impl SafeFloat {
+	pub fn try_from(value: f32) -> Option<Self> {
```
kvark: indentation is screwed up

src/safe_float.rs
kvark: let's use an even better thing: https://doc.rust-lang.org/stable/std/primitive.f32.html#method.to_bits

src/material.rs
```diff
@@ -182,7 +212,45 @@ impl Default for Pbr {
     }
 }
 
+#[doc(hidden)]
```
alteous: `pub(crate)`?

data/shaders/basic_vs.glsl
```diff
@@ -10,12 +11,11 @@ out vec4 v_Color;
 in vec4 i_World0;
 in vec4 i_World1;
 in vec4 i_World2;
-in vec4 i_World3;
 in vec4 i_Color;
 in vec4 i_UvRange;
 
 void main() {
-    mat4 m_World = mat4(i_World0, i_World1, i_World2, i_World3);
+    mat4 m_World = world_matrix(i_World0, i_World1, i_World2);
```
kvark: can we also move the `in vec4 i_World?` into the header?

data/shaders/world_matrix.glsl
```diff
@@ -0,0 +1,7 @@
+mat4 world_matrix(vec4 i_World0, vec4 i_World1, vec4 i_World2) {
```
kvark: I don't think it needs a header, tbh, since all you need is
`transpose(mat4(i_World0, i_World1, i_World2, vec4(0.0,0.0,0.0,1.0)))`

src/safe_float.rs
alteous: Tabs

src/material.rs
vitvakatu: No, can't expose private types in public `enum Material`

data/shaders/basic_vs.glsl
vitvakatu: If you think we need it, sure.

data/shaders/world_matrix.glsl
vitvakatu: Is transpose cheap operation?

src/material.rs
alteous: Hmm... I'm not happy about having two separate material kinds. Perhaps we should make `Material` opaque for now?

data/shaders/world_matrix.glsl
alteous: Yes.

vitvakatu:
Updated to gfx 16.3, rebased, ready for merge

kvark:
Thanks! A few more things I spotted.
Really happy that we got rid of those pesky unique IDs :)

---

Reviewed 6 of 17 files at r1, 4 of 10 files at r3, 7 of 9 files at r4, 7 of 8 files at r5.
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[Cargo.toml, line 27 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wrtQcSeP0k0pDK4Eg:-L1wrtQdgBgdds8PA_iC:b-woom3p) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/Cargo.toml#L27)):*
> ```TOML
> [dependencies]
> bitflags = "0.9"
> byteorder = "1.2"
> ```

this is not needed

---

*[examples/group.rs, line 135 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wsE5MFtTiRSwvyKLG:-L1wsE5MFtTiRSwvyKLH:be5uyhh) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/examples/group.rs#L135)):*
> ```Rust
> 
>     let timer = win.input.time();
>     println!("{}", cubes.len());
> ```

plz make it prettier a bit, describe what the number is for

---

*[src/material.rs, line 202 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wsSnuryrqYU9QLGZ3:-L1wsSnuryrqYU9QLGZ4:b-2x5mgd) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/material.rs#L202)):*
> ```Rust
> #[doc(hidden)]
> #[derive(Clone, Debug, PartialEq, Eq, Hash)]
> pub struct InternalPbr {
> ```

this goes against our "convention" (e.g. `Node` -> `NodeInternal`). How about `PbrInternal`?

---

*[src/material.rs, line 265 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1ws_ZM6tiee01HVGaH:-L1ws_ZM6tiee01HVGaI:b-9mrew3) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/material.rs#L265)):*
> ```Rust
> #[doc(hidden)]
> #[derive(Clone, Debug, PartialEq, Eq, Hash)]
> pub struct InternalPhong {
> ```

same here, let's be consistent

---

*[src/safe_float.rs, line 17 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wtXafqMS25z577li0:-L1wtXagLs-qQSyL-L_d:b-4xpyo4) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/safe_float.rs#L17)):*
> ```Rust
> 
>     pub fn new(value: f32) -> Self {
>         Self::try_from(value).expect("Illegal float value")
> ```

could be `From<f32>`

---

*[src/safe_float.rs, line 28 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wtcLtrk1YQW5Y-3ln:-L1wtcLtrk1YQW5Y-3lo:b-63a7gh) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/safe_float.rs#L28)):*
> ```Rust
>         state: &mut H,
>     ) {
>         self.0.to_bits().hash(state);
> ```

 wonder if it's going to be simpler to store it as `u32` and auto-derive `Hash` and some others

---

*[src/texture.rs, line 26 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wto3Dfo4a3NrMHo5b:-L1wto3Dfo4a3NrMHo5c:bz351g6) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/texture.rs#L26)):*
> ```Rust
> 
> impl<T> Hash for Texture<T> {
>     fn hash<H: Hasher>(
> ```

can this impl be auto-derived?

---

*[src/factory/mod.rs, line 263 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wu3C6ZTPbgjdC-RJ1:-L1wu3C6ZTPbgjdC-RJ2:b80l28r) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/factory/mod.rs#L263)):*
> ```Rust
>                 GpuData {
>                     slice,
>                     vertices: vbuf.clone(),
> ```

why is `clone()` needed here?

---

*[src/factory/mod.rs, line 305 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wu8F4vntJM-me7vQT:-L1wu8F4vntJM-me7vQU:bvkhquo) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/factory/mod.rs#L305)):*
> ```Rust
>         };
>         let instances = self.create_instance_buffer();
>         let material = material.into();
> ```

can we do this in-place below, like the old code did?

---

*[src/factory/mod.rs, line 368 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wuKacaxMsXPgxT8Sp:-L1wuKacaxMsXPgxT8Sq:btfmvqz) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/factory/mod.rs#L368)):*
> ```Rust
>                 instances,
>                 instance_cache_key: Some(InstanceCacheKey {
>                     material: material.clone(),
> ```

does it need a `clone()` here?

---

*[src/factory/mod.rs, line 889 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wuTWKvJk9oMNfnmn0:-L1wuTWKvJk9oMNfnmn1:b80l28r) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/factory/mod.rs#L889)):*
> ```Rust
>                         GpuData {
>                             slice,
>                             vertices: vbuf.clone(),
> ```

why is `clone()` needed here?

---

*[src/render/mod.rs, line 28 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wuZ5uNt6FQFrt5tRW:-L1wuZ5uNt6FQFrt5tRX:b-hy14ia) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L28)):*
> ```Rust
> pub use self::back::Factory as BackendFactory;
> pub use self::back::Resources as BackendResources;
> use self::pso_data::PsoData;
> ```

let's move it out of `pub use` section

---

*[src/render/mod.rs, line 214 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wuejaPTRQqNyMJLOZ:-L1wuejbSkdDPs7i-25P:b-tzhl7r) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L214)):*
> ```Rust
>         uv_range: [f32; 4],
>         param: f32,
>     ) -> Instance {
> ```

`Self`

---

*[src/render/mod.rs, line 216 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wujrsXd7XOlqyiwbg:-L1wujrsXd7XOlqyiwbh:b-5x2ggf) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L216)):*
> ```Rust
>     ) -> Instance {
>         Instance {
>             world0: mx_world[0],
> ```

hmm, shouldn't it be transposed here?

---

*[src/render/mod.rs, line 220 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wv-6LgprT8gmEahc3:-L1wv-6LgprT8gmEahc4:bcm5nyw) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L220)):*
> ```Rust
>             world2: mx_world[2],
>             color: {
>                 let rgb = color::to_linear_rgb(color);
> ```

we could make this helper more useful by adding a parameter for `W` and returning `[f32; 4]` right away. Perfectly fine for a follow-up :)

---

*[src/render/mod.rs, line 229 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wv8E5_7mAghRgynq8:-L1wv8E6mrkQXvbsZb2W:b-tzhl7r) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L229)):*
> ```Rust
> 
>     #[inline]
>     fn pbr(mx_world: [[f32; 4]; 4]) -> Instance {
> ```

`Self`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150#-:-L1wv_bPGNDZDygYdCwq:b-pj4rmv)*

vitvakatu:
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[Cargo.toml, line 27 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wrtQcSeP0k0pDK4Eg:-L1wzKJyE8q8t9EEZPCZ:b6p05rc) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/Cargo.toml#L27)):*

this is not needed
</blockquote></details>

Oh, rebasing issue. Nicely spotted!

---

*[src/safe_float.rs, line 17 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wtXafqMS25z577li0:-L1wzXsccZOM9XlN8is9:b-rflmfx) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/safe_float.rs#L17)):*

could be `From<f32>`
</blockquote></details>

No, `From` must not panic

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

vitvakatu:
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[Cargo.toml, line 27 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wrtQcSeP0k0pDK4Eg:-L1x0BGsK1KTxn7kF9yo:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/Cargo.toml#L27)):*
<details><summary><i>Previously, vitvakatu (Ilya Bogdanov) wrote…</i></summary><blockquote>

Oh, rebasing issue. Nicely spotted!
</blockquote></details>

Done.

---

*[examples/group.rs, line 135 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wsE5MFtTiRSwvyKLG:-L1x0MSYSSyVwZG31Wu5:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/examples/group.rs#L135)):*

plz make it prettier a bit, describe what the number is for
</blockquote></details>

Done.

---

*[src/material.rs, line 202 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wsSnuryrqYU9QLGZ3:-L1x0aewM0eg8jugw3T9:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/material.rs#L202)):*

this goes against our "convention" (e.g. `Node` -> `NodeInternal`). How about `PbrInternal`?
</blockquote></details>

Done.

---

*[src/material.rs, line 265 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1ws_ZM6tiee01HVGaH:-L1x0bM-KH4m2uJVzjRf:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/material.rs#L265)):*

same here, let's be consistent
</blockquote></details>

Done.

---

*[src/safe_float.rs, line 28 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wtcLtrk1YQW5Y-3ln:-L1x0dJ-riRsl_wOOxj7:b-w6kvw) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/safe_float.rs#L28)):*

 wonder if it's going to be simpler to store it as `u32` and auto-derive `Hash` and some others
</blockquote></details>

I prefer to leave it for now

---

*[src/texture.rs, line 26 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wto3Dfo4a3NrMHo5b:-L1x11GRsz7C032_nZId:b-idama0) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/texture.rs#L26)):*

can this impl be auto-derived?
</blockquote></details>

No, because of type parameter

---

*[src/factory/mod.rs, line 263 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wu3C6ZTPbgjdC-RJ1:-L1x16GYGW9vxw4yifzw:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/factory/mod.rs#L263)):*

why is `clone()` needed here?
</blockquote></details>

Done.

---

*[src/factory/mod.rs, line 305 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wu8F4vntJM-me7vQT:-L1x17OQzUdpFnSFjWsM:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/factory/mod.rs#L305)):*

can we do this in-place below, like the old code did?
</blockquote></details>

Done.

---

*[src/factory/mod.rs, line 368 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wuKacaxMsXPgxT8Sp:-L1x1Qsxnb_884FIYkXT:b2nhswy) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/factory/mod.rs#L368)):*

does it need a `clone()` here?
</blockquote></details>

Yes

---

*[src/factory/mod.rs, line 889 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wuTWKvJk9oMNfnmn0:-L1x1emM1n9gfM3WnhMn:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/factory/mod.rs#L889)):*

why is `clone()` needed here?
</blockquote></details>

Done.

---

*[src/render/mod.rs, line 28 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wuZ5uNt6FQFrt5tRW:-L1x1oPtb5rpMewJ7G2F:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L28)):*

let's move it out of `pub use` section
</blockquote></details>

Done.

---

*[src/render/mod.rs, line 214 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wuejaPTRQqNyMJLOZ:-L1x1sQ3EW1-W_3A6_Tx:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L214)):*

`Self`
</blockquote></details>

Done.

---

*[src/render/mod.rs, line 216 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wujrsXd7XOlqyiwbg:-L1x1tnfUoc9Z4_CNcXM:brbn7r2) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L216)):*

hmm, shouldn't it be transposed here?
</blockquote></details>

I assume that it's already transposed, so not

---

*[src/render/mod.rs, line 220 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wv-6LgprT8gmEahc3:-L1x24eucxVT62loqOMO:b-b87xf6) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L220)):*

we could make this helper more useful by adding a parameter for `W` and returning `[f32; 4]` right away. Perfectly fine for a follow-up :)
</blockquote></details>

Will cover in follow-up

---

*[src/render/mod.rs, line 229 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wv8E5_7mAghRgynq8:-L1x2938EwJMLPAXTcky:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L229)):*

`Self`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

kvark:
Reviewed 5 of 5 files at r6.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/texture.rs, line 26 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wto3Dfo4a3NrMHo5b:-L1x3mbPde86BTCcxunK:b-sl5z5t) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/texture.rs#L26)):*
<details><summary><i>Previously, vitvakatu (Ilya Bogdanov) wrote…</i></summary><blockquote>

No, because of type parameter
</blockquote></details>

Will be possible with `derivative` crate, like how gfx-rs does does this. (for a follow-up, but please leave a TODO comment there)

---

*[src/render/mod.rs, line 216 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wujrsXd7XOlqyiwbg:-L1x4aLnP2lTM46I3x1m:b-t7hsc0) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L216)):*
<details><summary><i>Previously, vitvakatu (Ilya Bogdanov) wrote…</i></summary><blockquote>

I assume that it's already transposed, so not
</blockquote></details>

I see. Let's either have it reflected in the parameter name, or even use the strongly typed `mint::RowMatrix` there

---

*[src/render/mod.rs, line 220 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wv-6LgprT8gmEahc3:-L1x4JcrMhQYOGtNvQQb:b-qmz2dw) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L220)):*
<details><summary><i>Previously, vitvakatu (Ilya Bogdanov) wrote…</i></summary><blockquote>

Will cover in follow-up
</blockquote></details>

thanks! please add a TODO comment

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

vitvakatu:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/texture.rs, line 26 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wto3Dfo4a3NrMHo5b:-L1x616wuy62iJuD01CY:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/texture.rs#L26)):*

Will be possible with `derivative` crate, like how gfx-rs does does this. (for a follow-up, but please leave a TODO comment there)
</blockquote></details>

Done.

---

*[src/render/mod.rs, line 216 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wujrsXd7XOlqyiwbg:-L1x7ZNhxE94mmguHUG9:b-7wfw3i) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L216)):*

I see. Let's either have it reflected in the parameter name, or even use the strongly typed `mint::RowMatrix` there
</blockquote></details>

Good advice, found issue in shadow rendering. Done.

---

*[src/render/mod.rs, line 220 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wv-6LgprT8gmEahc3:-L1x7xw3xA5XrxgGE4De:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L220)):*

thanks! please add a TODO comment
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

kvark:
Reviewed 2 of 2 files at r7.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/render/mod.rs, line 216 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wujrsXd7XOlqyiwbg:-L1x9rMFWMPAZTFYZEpG:b62ytv0) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L216)):*
<details><summary><i>Previously, vitvakatu (Ilya Bogdanov) wrote…</i></summary><blockquote>

Good advice, found issue in shadow rendering. Done.
</blockquote></details>

extra nit: no need to reflect it no both the name and the type. Only the type is clear enough, and we can have a shorter name ;)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

vitvakatu:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/render/mod.rs, line 216 at r5](https://reviewable.io:443/reviews/three-rs/three/150#-L1wujrsXd7XOlqyiwbg:-L1xBxy4aFUu4daa20El:b-896fix) ([raw file](https://github.com/three-rs/three/blob/941c747836b7f4f05450e4b8dbd696e7478a9240/src/render/mod.rs#L216)):*

extra nit: no need to reflect it no both the name and the type. Only the type is clear enough, and we can have a shorter name ;)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

kvark:
Reviewed 1 of 1 files at r8.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

kvark:
Reviewed 7 of 7 files at r9.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/texture.rs, line 4 at r9](https://reviewable.io:443/reviews/three-rs/three/150#-L1xURUyjVKd4cDqIxXo:-L1xURUyjVKd4cDqIxXp:b-f5a4m1) ([raw file](https://github.com/three-rs/three/blob/6f2f20ce4fc640e22239647238655cff12f5740d/src/texture.rs#L4)):*
> ```Rust
> use render::BackendResources;
> use std::path::Path;
> use util;
> ```

I typically order imports into groups: dependencies, this crate, std. Not a big deal though, I assume rustfmt will overwrite this?

---

*[src/texture.rs, line 21 at r9](https://reviewable.io:443/reviews/three-rs/three/150#-L1xUk9pXmC5ZXLLULmQ:-L1xUk9pXmC5ZXLLULmR:b-6farf2) ([raw file](https://github.com/three-rs/three/blob/6f2f20ce4fc640e22239647238655cff12f5740d/src/texture.rs#L21)):*
> ```Rust
>     sampler: h::Sampler<BackendResources>,
>     total_size: [u32; 2],
>     #[derivative(Hash(hash_with = "util::hash_f32_slice"))] tex0: [f32; 2],
> ```

let's split in 2 lines, like everywhere else

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

vitvakatu:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/texture.rs, line 4 at r9](https://reviewable.io:443/reviews/three-rs/three/150#-L1xURUyjVKd4cDqIxXo:-L1xXJWxkX0ye_I-G5qk:b-896fix) ([raw file](https://github.com/three-rs/three/blob/6f2f20ce4fc640e22239647238655cff12f5740d/src/texture.rs#L4)):*

I typically order imports into groups: dependencies, this crate, std. Not a big deal though, I assume rustfmt will overwrite this?
</blockquote></details>

Done.

---

*[src/texture.rs, line 21 at r9](https://reviewable.io:443/reviews/three-rs/three/150#-L1xUk9pXmC5ZXLLULmQ:-L1xXR6MrYtGjhsDUu_Q:b8lx6pu) ([raw file](https://github.com/three-rs/three/blob/6f2f20ce4fc640e22239647238655cff12f5740d/src/texture.rs#L21)):*

let's split in 2 lines, like everywhere else
</blockquote></details>

Oh, it's rustfmt. Should I disable formatting for this struct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/three-rs/three/150)*

kvark:
🐑 it 💯 🥇 
bors r+

alteous:
:shipit:

PR #151, opened 2017-12-19
[WIP] Animations part 2 of 2

alteous:
This PR implements the second part of the animation system targeting mesh skinning, mesh morphing, and the [omitted features](https://github.com/three-rs/three/pull/103) from part 1.

Implemented features

* Set of Action, Mixer, and Clip structs, inspired by the three.js animation system.
* Keyframe targets for simple binding pairs.
  * (Position, Vector3)
  * (Orientation, Quaternion),
  * (Orientation, Euler)
  * (Scale, Scalar)
  * (Weight, Scalar)
* Linear interpolation between keyframes.
* Support for node animations in the glTF loader.
* Two loop modes, namely Once and Repeat.
* Mesh skinning for meshes with a small (≤ 20) number of joints per-vertex.

WIP features
* Mesh skinning for meshes with a large number of joints per-vertex.
* Morph targets.
* Implementation of Discrete, Cubic, and CatmullRom interpolation modes.
* Implementation of PingPong loop mode.


Cargo.toml
```diff
@@ -23,15 +23,16 @@ default = ["opengl"]
...
 bitflags = "0.9"
 cgmath = { version = "0.15", features = ["mint"] }
 froggy = "0.4.4"
 genmesh = "0.5"
 gfx = "0.16"
 gfx_glyph = "0.3"
-gltf = { features = ["names"], version = "0.9.2" }
-gltf-importer = { features = ["names"], version = "0.9.2" }
-gltf-utils = { features = ["names"], version = "0.9.2" }
+gltf = { features = ["names"], version = "0.10.0", path = "../gltf" }
```
alteous: Need to release `gltf 0.10.0` first!

data/shaders/locals.glsl
```diff
@@ -1,6 +1,7 @@
-uniform b_Locals {
-    mat4 u_World;
+layout(std140) uniform b_Locals {
     vec4 u_Color;
     vec4 u_MatParams;
     vec4 u_UvRange;
+    mat4 u_World;
+    mat4 u_JointMatrix[20];
```
alteous: Needs to be passed as a texture instead.

src/render/mod.rs
```diff
@@ -92,13 +93,35 @@ gfx_defines! {
...
+        weights: [f32; 4] = "a_Weight",
     }
 
     constant Locals {
-        mx_world: [[f32; 4]; 4] = "u_World",
         color: [f32; 4] = "u_Color",
         mat_params: [f32; 4] = "u_MatParams",
         uv_range: [f32; 4] = "u_UvRange",
+        mx_world: [[f32; 4]; 4] = "u_World",
+        joint_matrix_0: [[f32; 4]; 4] = "u_JointMatrix[0]",
```
alteous: 😇

src/render/mod.rs
```diff
@@ -701,18 +744,57 @@ impl Renderer {
...
 
+            let mut joint_matrices = [Matrix4::identity(); 20];
+            if let &Some(ref object) = skeleton {
+                let data = match hub.get(object).sub_node {
+                    hub::SubNode::Skeleton(ref data) => data,
+                    _ => unreachable!(),
+                };
+                let (bones, ibms) = (&data.bones, &data.inverses);
+                if bones.len() > 20 {
+                    eprintln!("Joint limit exceeded ({}/20)", bones.len());
```
alteous: Temporary, obviously.

src/render/mod.rs
```diff
@@ -701,18 +744,57 @@ impl Renderer {
...
+                    ::std::process::exit(1);
+                }
+                for (i, (bone, ibm)) in izip!(bones.iter(), ibms.iter()).enumerate() {
```
alteous: Let me know if this can be cleaned up a bit.

src/mesh.rs
```diff
@@ -114,6 +137,24 @@ impl Mesh {
...
+        skeleton: Skeleton,
+    ) {
+        let msg = Operation::SetSkeleton(skeleton);
+        let _ = self.object.tx.send((self.object.node.downgrade(), msg));
+    }
+
+    /// Bind a set of morph targets to the mesh.
+    pub fn set_targets(
+        &mut self,
+        targets: ArrayVec<[Target; MAX_TARGETS]>,
```
alteous: Ideally we'd take an `Iterator<Item = Target>` here but since there's a hard limit on the number of targets I wasn't so sure.

src/mesh.rs
```diff
@@ -1,12 +1,35 @@
...
+    Position,
+
+    /// Normal displacements.
+    Normal,
+
+    /// Tangent displacements.
+    Tangent,
+}
+
+pub enum Weight {
```
alteous: Missing docs.

src/factory/mod.rs
```diff
@@ -216,13 +287,32 @@ impl Factory {
...
         };
-        izip!(position_iter, normal_iter, tangent_iter, uv_iter)
-            .map(|(position, normal, tangent, tex_coord)| {
+        let joints_iter = if shape.joints.is_empty() {
+            Either::Left(iter::repeat([0.0, 0.0, 0.0, 0.0]))
+        } else {
+            Either::Right(shape.joints.iter().cloned())
+        };
+        let weights_iter = if shape.weights.is_empty() {
+            Either::Left(iter::repeat([1.0, 0.0, 0.0, 0.0]))
```
alteous: N.B!

src/factory/load_gltf.rs
```diff
@@ -84,42 +152,44 @@ impl super::Factory {
...
             is_basic_material = false;
-            self.load_gltf_texture(&t, buffers, base)
+            textures[t.as_ref().index()].clone()
         });
         let (base_color_factor, base_color_alpha) = {
             let x = pbr.base_color_factor();
             (color::from_linear_rgb([x[0], x[1], x[2]]), x[3])
         };
-        if is_basic_material {
+        let material = if false {// is_basic_material {
```
alteous: Temporary.

src/factory/load_gltf.rs
```diff
@@ -182,239 +258,221 @@ impl super::Factory {
             } else {
                 Vec::new()
             };
+            let joints = if let Some(iter) = primitive.joints_u16(0, buffers) {
```
alteous: Broken in one of the recent versions of `gltf_utils`.

src/factory/load_gltf.rs
```diff
@@ -182,239 +258,221 @@ impl super::Factory {
...
+                // This animation does not correspond to any loaded node.
+                None => continue,
+            };
+            let input = sampler.input();
+            let output = sampler.output();
+            let interpolation = match sampler.interpolation() {
+                Linear => animation::Interpolation::Linear,
+                Step => animation::Interpolation::Discrete,
+                CubicSpline => animation::Interpolation::Cubic,
+                CatmullRomSpline => animation::Interpolation::CatmullRom,
```
alteous: Should we try to support this anyway?

src/factory/load_gltf.rs
```diff
@@ -182,239 +258,221 @@ impl super::Factory {
...
+    meshes: &VecMap<Vec<Mesh>>,
+    skeletons: &[Skeleton],
+) -> Vec<Mesh> {
+    let mut instances = Vec::new();
+    for node in gltf.nodes() {
+        if let Some(ref group) = heirarchy.get(node.index()) {
+            if let Some(mesh) = node.mesh() {
+                let mut primitives = meshes[mesh.index()]
+                    .iter()
+                    .map(|template| factory.mesh_instance(template))
```
alteous: @vitvakatu How does this align with the new auto-instancing?

src/factory/load_gltf.rs
vitvakatu: I don't think we should.

src/factory/load_gltf.rs
vitvakatu: Well, if meshes have the same material - they will be auto-grouped and rendered as instanced.

src/mesh.rs
vitvakatu: Uhm, maybe we should not expose `ArrayVec` to user?

src/factory/mod.rs
```diff
@@ -665,7 +767,7 @@ impl Factory {
         let size = images[0].dimensions().0;
         let kind = t::Kind::Cube(size as t::Size);
         let (_, view) = factory
-            .create_texture_immutable_u8::<gfx::format::Srgba8>(kind, &data)
+            .create_texture_immutable_u8::<gfx::format::Srgba8>(kind, gfx::texture::Mipmap::Allocated, &data)
```
alteous: Is this correct?

src/mesh.rs
alteous: Not sure; it's up for debate. An alternative is to take `IntoIterator`.

src/render/mod.rs
```diff
@@ -418,17 +424,32 @@ impl Renderer {
         use gfx::texture as t;
         let (window, device, mut gl_factory, out_color, out_depth) = gfx_window_glutin::init(builder, context, event_loop);
         let (_, srv_white) = gl_factory
-            .create_texture_immutable::<gfx::format::Rgba8>(t::Kind::D2(1, 1, t::AaMode::Single), &[&[[0xFF; 4]]])
+            .create_texture_immutable::<gfx::format::Rgba8>(t::Kind::D2(1, 1, t::AaMode::Single), t::Mipmap::Provided, &[&[[0xFF; 4]]])
```
alteous: Is this correct?

src/factory/mod.rs
alteous: This isn't as special as I originally thought. The new behaviour is to take the weights `[1.0, 1.0, 1.0, 1.0]` with joints `[0.0, 0.0, 0.0, 0.0]` so that an identity matrix is computed when there is no skeleton (i.e. the default joint transform buffer is used).

alteous:
We're slowing getting there...

One notable change is the way morph targets are being computed. I've introduced generic 'displacement' vertex attributes which influence `a_Position`, `a_Normal`, and `a_Tangent` subject to the values of `u_DisplacementContributions`.

The snippet below taken from `data/shaders/pbr_vs.glsl` should explain a bit more:
```glsl
// Toggles displacement contributions to `a_Position/a_Normal/a_Tangent`.
struct DisplacementContribution {
    // position: 1.0 if morph target weights should influence a_Position
    // normal: 1.0 if morph target weights should influence a_Normal
    // tangent: 1.0 if morph target weights should influence a_Tangent
    // weight: The weight to be applied.
    float position, normal, tangent, weight;
};
```

PR #153, opened 2017-12-28
Fix URL for car screenshot in README.md

### LINKED ISSUE #146: One screenshot from readme is missing after test_data moving.

randomPoison:
Fixes #146

I opted to use the fully-qualified URL (including the commit hash) for the image, which should ensure that the link stays valid even if the image is moved or deleted in a later commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ If you a looking for something simple to prototype 3D graphics with, you found i
 ## Screenshots
 
 ![Aviator](examples/aviator/shot.png)
-![CarObj](test_data/obj-car.png)
+![CarObj](https://raw.githubusercontent.com/three-rs/example-data/5d821bc9647a8db26888f8dd286f318eaabd2a52/obj-car.png)
 
 ## Motivation and Goals
 


kvark:
Thank you for the PR!
I'm not certain about the fully qualified URL. It's pointing to our own repo, after all, so if something goes wrong with it, this very repo may as well be down.

randomPoison:
I'm not sure I understand what you're saying. Are you referring to the fact that I used the `raw.githubusercontent.com` URL instead of a relative path? I don't believe a relative path would work, since the file is now in a submodule (though I could always move the file into this repo, if that's the preferred solution).

kvark:
I mean, why not just use `https://github.com/three-rs/example-data/raw/master/obj-car.png` link?

randomPoison:
Okay, so you're primarily concerned with using the `master` branch link vs using the specific commit hash? My thinking was that if the image is moved or deleted from the `example-data` repo, then the URL that uses the `master` branch will break, whereas the URL using the commit hash will always work forever. Sure, you control both repos so you can always update the URL if the file moves, but using the commit hash means you don't *have to* update the URL immediately. That said, if you'd prefer to use the `master` branch in the link, then I'll update the PR accordingly.

Also, regarding `github.com` vs `raw.githubusercontent.com`: `https://github.com/three-rs/example-data/raw/master/obj-car.png` redirects to `https://raw.githubusercontent.com/three-rs/example-data/master/obj-car.png`, since `raw.githubusercontent.com` is the site that static content gets served from. I don't know if that's the part you care about, but I figured linking directly to `raw.githubusercontent.com` was the better way of handling it, since that's where the image is *actually* coming from.

vitvakatu:
I don't think it worth discussion. We can always fix (or replace) link later if something goes really bad.

kvark:
Agreed

Bors r+

PR #155, opened 2018-01-03
Node tree rewrite

### LINKED ISSUE #44: Node children lifetime

kvark:
Fixes #44

Core of the change is switching relationships from "child->parent" to "parent->first_child->sibling", which gives us the following benefits:
  - clean children lifetimes, as expected by users
    - no need to store all the meshes returned by gltf/obj loaders: perhaps, we can follow-up with a PR that allows querying meshes by name from their parents, so that we don't even return all the meshes directly in the loader API
  - nice `Group::add(&something)`, like Three-js does
  - make `Scene` truly unique and node-less
  - no need for unique scene IDs
    ~~- technically, objects can now be present in multiple scenes at once~~
  - immutable node data upon traversal (!):
    - we just compute the world visibility/transform on the fly
    - thus, no need for a separate `update_transforms` step
    - uses pretty iterators ;)
    - as a downside, not as straightforward to reflect the world info to the user (see TODO)

Has other changes making things nicer:
  - refactored frame render code
  - no `&mut self` on objects
  - idiomatic hub indexing
  - safer group/camera/text/audio spawning (internal)
  - more robust message handling (panic on unprocessed/unexpected messages)
  - can debug print the `Operation`

TODO:
- [x] check/fix docs
- [x] compute world transform/visibility for the user if needed (is it needed?)
- [x] provide API to detach from parent
- [x] reviews!
- [x] profile the cost of recomputing the world transforms

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ cgmath = { version = "0.15", features = ["mint"] }
 derivative = "1.0"
 froggy = "0.4.4"
 genmesh = "0.5"
-gfx = "0.16"
+gfx = "0.16.3"
 gfx_glyph = "0.7"
 gltf = { features = ["names"], version = "0.9.2" }
 gltf-importer = { features = ["names"], version = "0.9.2" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ exclude = ["doc", "bors.toml", ".travis.yml", "test_data"]
 
 [lib]
 
+[profile.release]
+debug = true
+
 [features]
 default = ["opengl"]
 "opengl" = ["gfx_device_gl", "gfx_window_glutin", "glutin"]
--- a/examples/group.rs
+++ b/examples/group.rs
@@ -102,9 +102,19 @@ fn create_cubes(
     list
 }
 
-const COLORS: [three::Color; 6] = [0xffff80, 0x8080ff, 0x80ff80, 0xff8080, 0x80ffff, 0xff80ff];
-
-const SPEEDS: [f32; 6] = [0.7, -1.0, 1.3, -1.6, 1.9, -2.2];
+struct LevelDesc {
+    color: three::Color,
+    speed: f32, // in radians per second
+}
+const LEVELS: &[LevelDesc] = &[
+    LevelDesc { color: 0xffff80, speed: 0.7 },
+    LevelDesc { color: 0x8080ff, speed: -1.0 },
+    LevelDesc { color: 0x80ff80, speed: 1.3 },
+    LevelDesc { color: 0xff8080, speed: -1.6 },
+    LevelDesc { color: 0x80ffff, speed: 1.9 },
+    LevelDesc { color: 0xff80ff, speed: -2.2 },
+    //LevelDesc { color: 0x8080ff, speed: 2.5 },
+];
 
 fn main() {
     let mut win = three::Window::new("Three-rs group example");
@@ -117,11 +127,14 @@ fn main() {
     light.set_position([0.0, -10.0, 10.0]);
     win.scene.add(&light);
 
-    let materials: Vec<_> = COLORS
+    let materials = LEVELS
         .iter()
-        .map(|&color| three::material::Lambert { color, flat: false })
-        .collect();
-    let levels: Vec<_> = SPEEDS.iter().map(|&speed| Level { speed }).collect();
+        .map(|l| three::material::Lambert { color: l.color, flat: false })
+        .collect::<Vec<_>>();
+    let levels = LEVELS
+        .iter()
+        .map(|l| Level { speed: l.speed })
+        .collect::<Vec<_>>();
     let mut cubes = create_cubes(&mut win.factory, &materials, &levels);
     win.scene.add(&cubes[0].group);
 
--- a/examples/group.rs
+++ b/examples/group.rs
@@ -27,7 +27,7 @@ fn create_cubes(
     }
 
     let root = {
-        let mut group = factory.group();
+        let group = factory.group();
         let mesh = factory.mesh(geometry.clone(), materials[0].clone());
         group.set_position([0.0, 0.0, 1.0]);
         group.set_scale(2.0);
@@ -78,7 +78,7 @@ fn create_cubes(
     while let Some(next) = stack.pop() {
         for child in &children {
             let mat = materials[next.mat_id].clone();
-            let mut cube = Cube {
+            let cube = Cube {
                 group: factory.group(),
                 mesh: factory.mesh_instance_with_material(&list[0].mesh, mat),
                 level_id: next.lev_id,
@@ -110,10 +110,10 @@ fn main() {
     let mut win = three::Window::new("Three-rs group example");
     win.scene.background = three::Background::Color(0x204060);
 
-    let mut cam = win.factory.perspective_camera(60.0, 1.0 .. 100.0);
+    let cam = win.factory.perspective_camera(60.0, 1.0 .. 100.0);
     cam.look_at([-1.8, -8.0, 7.0], [0.0, 0.0, 3.5], None);
 
-    let mut light = win.factory.point_light(0xffffff, 1.0);
+    let light = win.factory.point_light(0xffffff, 1.0);
     light.set_position([0.0, -10.0, 10.0]);
     win.scene.add(&light);
 
--- a/examples/group.rs
+++ b/examples/group.rs
@@ -28,10 +28,10 @@ fn create_cubes(
 
     let root = {
         let mut group = factory.group();
-        let mut mesh = factory.mesh(geometry.clone(), materials[0].clone());
+        let mesh = factory.mesh(geometry.clone(), materials[0].clone());
         group.set_position([0.0, 0.0, 1.0]);
         group.set_scale(2.0);
-        mesh.set_parent(&group);
+        group.add(&mesh);
         Cube {
             group,
             mesh,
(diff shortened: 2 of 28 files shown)


src/hub.rs
```diff
@@ -185,24 +148,55 @@ impl Hub {
...
+                    }
+                    _ => unreachable!()
                 },
+            };
+        }
+
+        for (child_ptr, sibling) in deferred_sibling_updates {
+            let child = &mut self.nodes[&child_ptr];
+            if child.next_sibling.is_some() {
+                error!("Attaching a child that still has an old parent, discarding siblings");
```
vitvakatu: How we can change parent now?

kvark: Oh right, I forgot about this :) Now added to TODO list for this PR.
Basically, we can either merge the old siblings with the new ones (like `Scene::add` does), or require the user to call `Group::remove(&item)` before adding to another group. Either way, the `remove` needs to be added.

src/hub.rs
```diff
@@ -185,24 +148,55 @@ impl Hub {
...
-                    *mat = material;
                 },
-                Operation::SetTexelRange(base, size) => if let SubNode::Visual(ref mut material, _) = node.sub_node {
-                    match *material {
-                        material::Material::Sprite(ref mut params) => params.map.set_texel_range(base, size),
-                        _ => panic!("Unsupported material for texel range request"),
+                Operation::AddChild(child_ptr) => match node.sub_node {
+                    SubNode::Group { ref mut first_child } => {
+                        let sibling = mem::replace(first_child, Some(child_ptr.clone()));
+                        deferred_sibling_updates.push((child_ptr, sibling));
```
alteous: Interesting.

src/hub.rs
```diff
@@ -240,40 +234,83 @@ impl Hub {
...
         mesh: &DynamicMesh,
     ) {
-        match self.get_mut(&mesh).sub_node {
+        match self[mesh].sub_node {
             SubNode::Visual(_, ref mut gpu_data) => gpu_data.pending = Some(mesh.dynamic.clone()),
             _ => unreachable!(),
         }
     }
+
+    pub(crate) fn walk(&self, base: &Option<NodePointer>) -> TreeWalker {
```
alteous: 👍

src/node.rs
```diff
@@ -72,28 +74,10 @@ impl From<TransformInternal> for Transform {
...
-    pub transform: Transform,
-    /// World transform (relative to the world's origin).
-    pub world_transform: Transform,
     /// Is `Node` visible by cameras or not?
     pub visible: bool,
-    /// The same as `visible`, used internally.
-    pub world_visible: bool,
+    /// Relative to parent transform.
+    pub transform: Transform,
     /// Material in case this `Node` has it.
```
alteous: Was this ever useful?

src/render/mod.rs
```diff
@@ -654,13 +653,9 @@ impl Renderer {
...
             let p: [[f32; 4]; 4] = camera.matrix(self.aspect_ratio()).into();
-            let node = &hub.nodes[&camera.object.node];
-            let w = match node.scene_id {
-                Some(id) if Some(id) == scene_id => node.world_transform,
-                Some(_) => panic!("Camera does not belong to this scene"),
-                None => node.transform,
-            };
-            let mx_view = Matrix4::from(w.inverse_transform().unwrap());
+            let node = &hub[&camera];
+            let world_transform = node.transform; //TODO!!!
```
alteous: What's this for?

src/scene.rs
```diff
@@ -4,15 +4,13 @@ use object;
...
-use object::Object;
+use node::{Node, NodePointer};
+use object::{Base, Object};
 use texture::{CubeMap, Texture};
 
+use std::mem;
 use std::sync::MutexGuard;
 
-/// Unique identifier for a scene.
-pub type Uid = usize;
```
alteous: 👍

src/hub.rs
kvark: borrow checker...

src/node.rs
kvark: wait, why is `Material` here? O_O

alteous: Why not?!

kvark: Ah, this is user-exposed node, not the internal one.
Yeah, let it be. The whole reflection thing (with `sync`) I'd consider non-idiomatic, so we shouldn't be too eager to optimize it.

src/render/mod.rs
kvark: TODO: compute the world transform for the camera, which is expected here

kvark:
Alright, I rebased the code and implemented the missing bits.
Only the profiling is left, plus maybe a few more looks, esp in the new code.
Note: unfortunately, there is a bit of duplication atm between Group reacting to events and Scene implementation. I'm sure we can make it nicer.

kvark:
Here are the profiling results (note: I only used `callgrind_annotate`, so don't have that much insight at the moment):
  1. `descend` does show up near the top of the profile
  2. cause likely being large amount of copying (and a bit of `stack` reallocation). This should be partially solved by future improvements in Rust itself, some possible workarounds from our side, or caching the transforms at least.
  3. `group` examples with 20K instances runs OK on my U-series CPU, so I think we are fine for now.

Looks like I addressed all the action items from my side. Waiting for more reviews now (@alteous , @vitvakatu ).

kvark:
I believe in the long term our architecture should work like this:
- `Renderer` only walks the tree once, for each node
  - figures out what render bins does an object fall into (e.g.: main screen, shadow for light X, off-screen buffer, etc)
  - builds the GPU instance blocks for each bin (separately) and appends them to lists to be directly uploaded to GPU constant buffer memory
  - writes down GPU commands in the appropriate command buffers

TL;DR: there is no strong reason for `Renderer` to walk the tree multiple times. We can make rendering much faster on the CPU side, and tree traversal is not (going to be) a bottleneck.

examples/group.rs
```diff
@@ -102,28 +102,41 @@ fn create_cubes(
     list
 }
 
-const COLORS: [three::Color; 6] = [0xffff80, 0x8080ff, 0x80ff80, 0xff8080, 0x80ffff, 0xff80ff];
-
-const SPEEDS: [f32; 6] = [0.7, -1.0, 1.3, -1.6, 1.9, -2.2];
+struct LevelDesc {
+    color: three::Color,
+    speed: f32,
```
alteous: What is the unit of this? I assume radians/second?

Cargo.toml
```diff
@@ -28,7 +31,7 @@ cgmath = { version = "0.15", features = ["mint"] }
 derivative = "1.0"
 froggy = "0.4.4"
 genmesh = "0.5"
-gfx = "0.16"
+gfx = "0.16.3"
```
vitvakatu: Why?

src/hub.rs
```diff
@@ -240,40 +272,94 @@ impl Hub {
...
+        let mut walker = TreeWalker {
+            hub: self,
+            only_visible,
+            stack: Vec::with_capacity(default_stack_size),
+        };
+        walker.descend(base);
+        walker
+    }
+
+    pub(crate) fn walk(&self, base: &Option<NodePointer>) -> TreeWalker {
```
alteous: Nit: Perhaps `walk_visible`?

examples/shapes.rs
```diff
@@ -7,34 +7,40 @@ use three::Object;
...
         let material = three::material::Wireframe { color: 0xFF0000 };
         win.factory.mesh(geometry, material)
     };
     msphere.set_position([-3.0, 3.0, 0.0]);
-    msphere.set_parent(&win.scene);
+    win.scene.add(&msphere);
 
-    let mut mline = {
+    // test removal from scene
+    win.scene.remove(&mcyl);
```
vitvakatu: Is it supposed to be here?

src/hub.rs
vitvakatu: As in froggy, it's better to use shorter names.

Cargo.toml
kvark: Because your instancing changes require it

examples/group.rs
kvark: correct

examples/shapes.rs
kvark: yes - it's a small test for object removal. May not be the best place, but better than nothing

src/hub.rs
kvark: it's for internal use only anyhow. Most of the time we are only concerned about visible nodes

Cargo.toml
vitvakatu: I'm not a fan of strict version rules in Cargo. It will use the latest available version anyway.

src/hub.rs
vitvakatu: Yes, I remember your strategy :smile:
alt

Cargo.toml
kvark: I think your position is in the minority within the community then. Latest will be used on fresh checkout, if we don't check in `Cargo.lock`. But people who already have `Cargo.lock` will get breakage if you don't write the exact version in the `Cargo.toml`. Another case is - some crate that depends on you may say `gfx = "=0.16.2"` and it will try to build but will fail even after `cargo update`.

TL;DR: Cargo versioning rules are here to stay and be respected

kvark: Also, there is a proposal (which I supported) about making `cargo publish` checking out all the **minimal** compatible versions (as opposed to latest). Think about it ;)

vitvakatu: Okay, you convicted me

kvark:
bors r-
Almost forgot to fix the text!!!

kvark:
bors r=vitvakatu

PR #157, opened 2018-01-03
Enable multisampling support for Rasterizer

vitvakatu:
It's not trivial to make it adjustable by window builder settings, so is it okay to enable it by default?


vitvakatu:
MSAA can be requested from glutin via window builder. It works, I've checked.

bors r=kvark

Issue #156, opened 2018-01-03
Optimize compile times


kvark:
I noticed that we are spending multiple seconds compiling even some trivial changes. We need to investigate and see if something can be done about this. E.g. we could hide gltf stuff behind a (default) feature, or some derivatives, or use less generics, etc.

alteous:
I feel I must apologise for the atrocious compile time of `gltf`. It's become a bit of a joke. 😛

kvark:
@alteous would you be good with having `gltf` feature? I think it's not even have to be default.

alteous:
Yes, that's perfectly reasonable. Be aware though the documentation uses `gltf` in the examples.

kvark:
@alteous not sure, but perhaps the `animation.rs` documentation should provide non-gltf examples instead

alteous:
Although the compile time of `gltf` is really awful, it would be a shame to not have it as a default feature since it would limit the available 3D formats to just OBJ. Also, although glTF standardises PBR, we don't need to have the PBR shader with the glTF loader since in many cases we can fall back to the `Basic` material kind.

kvark:
I don't think having it as a feature would exclude anyone. Adding a simple `features = ["gltf"]` to Cargo seems simple enough. I see it a candidate for a feature because it's fairly self-contained and has it's own specific external dependencies and examples.

alteous:
I agree it should be a feature; I'm saying it should be enabled by default.

kvark:
Let me clarify my position, please:
  1. By no means I want to make the current glTF logic unavailable to the users - it's amazing, and I consider it a pillar of three-rs. Not that many projects feature high quality glTF + PBR + animation support.
  2. We both seem to agree that gltf needs to be gated a feature
  3. If the question raises "should this feature be default", then given all our equals (!), the answer should be "no". Users who don't care will get faster compile times (and run-times, given no need to link some shaders?). Users who do care - can easily switch it on. Unless there is some negative effect on the core function of the project, a feature shouldn't be default.
  4. Same can be said about `obj`, but it's way less code there.

alteous:
Reluctantly, I'm happy to let `gltf` be opt-in for the sake of compile times. I hesitate due to conflict of interest of wanting to see glTF as the 'standard' ("jpeg of 3D") runtime format for 3D opposed to other legacy and, frankly, inferior formats.

kvark:
[...]

PR #160, opened 2018-01-05
Add gltf-loader feature

alteous:
Addresses part of #156 .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,8 @@ exclude = ["doc", "bors.toml", ".travis.yml", "test_data"]
 
 [features]
 default = ["opengl"]
-"opengl" = ["gfx_device_gl", "gfx_window_glutin", "glutin"]
+gltf-loader = ["gltf", "gltf-importer", "gltf-utils"]
+opengl = ["gfx_device_gl", "gfx_window_glutin", "glutin"]
 
 [build-dependencies]
 includedir_codegen = "0.2.0"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -52,3 +53,11 @@ glutin = { version = "0.12", optional = true }
 env_logger = "0.4"
 notify = "4"
 rand = "0.3"
+
+[[example]]
```
kvark: I'm curious. Does registering only some of the examples here still allows running the others? I.e. have you tried running other examples with this change?

src/animation.rs
```diff
@@ -31,7 +31,7 @@
 //!
 //! First, we create a [`Mixer`] to play our animation.
 //!
-//! ```rust,no_run
+//! ```rust,no_run,ignore
```
kvark: why is `ignore` needed here? Perhaps, we need a TODO comment to rewrite those docs without the use of gtf-loader

Cargo.toml
alteous: Oops. Nope, will fix.

src/animation.rs
alteous: The TODOs would be difficult to notice.

```rust
//! ```rust,no_run,ignore
//! # // TODO: Rewrite test
//! ...
//! ```

alteous: Oh, perhaps just at the crate root instead?

Cargo.toml
kvark: ouch, too bad I removed them. Maybe you can just grab that from my changeset to not rewrite by hand

src/animation.rs
kvark: yep, that would do

kvark:
bors r-
wait, we need to make sure it's CI tested

kvark:
I suggest we add `--all-features` to `cargo test` while leaving `cargo build` untouched

alteous:
> I suggest we add --all-features to cargo test while leaving cargo build untouched

Add to where? Travis CI?

vitvakatu:
Yes

alteous:
Should be good to go now.

vitvakatu:
Looking good to me, merging
bors r+

PR #163, opened 2018-01-07
Add support for grabbing cursor

randomPoison:
Glutin/winit already supports grabbing the cursor, so I've added a new function `Window::set_cursor_state` that directly exposes the glutin/winit API. I've also re-exposed `glutin::CursorState`, in order to make it easy to use `set_cursor_state`.

I've tested this behavior on Windows and it seems to work as I'd expect, but it would be good to test it on Linux and MacOS, as well. There are a few open issues for winit related to cursor handling, but none that seem like they would drastically impact this functionality.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -12,6 +12,8 @@ use render::Renderer;
 use scene::Scene;
 use std::path::PathBuf;
 
+pub use glutin::CursorState;
+
 /// `Window` is the core entity of every `three-rs` application.
 ///
 /// It provides [user input](struct.Window.html#method.update),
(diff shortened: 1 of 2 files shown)


src/window.rs
```diff
@@ -253,6 +256,22 @@ impl Window {
...
+    /// Note that if you use [`CursorState::Grab`], you should use [`Input::mouse_delta_raw`] for
+    /// detecting mouse movement, as [`Input::mouse_delta`] will only report movement of the cursor
+    /// within the window.
+    ///
+    /// [`CursorState`]: enum.CursorState.html
+    /// [`CursorState::Grab`]: enum.CursorState.html#variant.Grab
+    /// [`Input::mouse_delta_raw`]: struct.Input.html#method.mouse_delta_raw
+    /// [`Input::mouse_delta`]: struct.Input.html#method.mouse_delta_raw
+    pub fn set_cursor_state(&self, state: CursorState) {
+        let _ = self.window.set_cursor_state(glutin::CursorState::Grab);
```
kvark: looks like this code ignores the `state` parameter

randomPoison: Womp womp, I totally messed that up. I've pushed a fix.

src/window.rs
```diff
@@ -242,6 +244,7 @@ impl Window {
     }
 
     /// Set cursor visibility
+    #[deprecated(note = "Use `set_cursor_state` instead")]
```
vitvakatu: Let's remove it! We're not caring much about backward compatibility at the moment.

randomPoison:
I've removed `show_cursor` and squashed this to a single commit. Let me know if there's anything else I should do!

PR #181, opened 2018-02-07
fix custom pipelines

tinaun:


vitvakatu:
Interesting solution...

src/render/mod.rs
```diff
@@ -428,12 +428,13 @@ impl PipelineStates {
         })
     }
 
-    pub(crate) fn pso_by_material(
-        &self,
-        material: &Material,
-    ) -> &BasicPipelineState {
+    pub(crate) fn pso_by_material<'a>(
```
kvark: Are the explicit lifetimes really needed here? I'd expect them to default to `&self` anyway

tinaun: they are. the default inference is

```
fn pso_by_material<'a, 'b>(&'a self, &'b Material) -> &'a BasicPipelineState
```
with 'b being an independent lifetime which could be less than 'a. this forces them to have the same lifetime

tinaun: at least, thats what rustc told me

kvark:
Alright, thank you!
bors r+

PR #184, opened 2018-02-22
glTF skinning and morphing animations

kvark:
This is a revival, rebase, and a substantial rewrite of #151

cc @alteous - if you want to check out how the bones are traversed without caching the world transforms

three-rs-skinning


data/shaders/pbr_vs.glsl
```diff
@@ -1,59 +1,101 @@
-
```
alteous: I'm not sure who removed this but beware of possible copyright violation.

src/render/mod.rs
```diff
@@ -592,7 +689,49 @@ impl Renderer {
...
+            use node::TransformInternal;
+
+            struct SkeletonTemp {
+                inverse_world_transform: TransformInternal,
+                cpu_buffer: Vec<[f32; 4]>,
+                gpu_buffer: gfx::handle::Buffer<BackendResources, [f32; 4]>,
+            }
+
+            let mut skeletons = Vec::new();
+            for w in hub.walk(&scene.first_child) {
```
alteous: 1

src/render/mod.rs
```diff
@@ -729,7 +868,7 @@ impl Renderer {
 
             for w in hub.walk(&scene.first_child) {
```
alteous: 2

src/render/mod.rs
```diff
@@ -789,39 +928,37 @@ impl Renderer {
         }
 
         for w in hub.walk(&scene.first_child) {
```
alteous: 3

kvark:
@alteous thanks for taking a look!

> I'm not sure who removed this but beware of possible copyright violation.

does the shader code match anything licensed? I thought you wrote it, and then I changed it more. If there are still lines under license, I'd like to rewrite them anyway.

> It seems you are traversing the scene graph (hub.walk) three times every frame. IIRC, you were pretty adamant about not traversing it more than once in the original PR.

You might have misunderstood me then. I said previously that **ideally** in some not far future we'd want to only traverse the graph once and collect all the render data into appropriate buckets for GPU. That doesn't in any way force this change to go crazy out of scope, especially since there were already multiple passes over the graph before it.

And that *ideal* scenario was brought up as an argument to not cache those world transforms (or anything, really, we are almost functional!). As you can see, bone transforms are fitting the scheme just fine.

alteous:
The majority of the shader code was taken from https://github.com/KhronosGroup/glTF-WebGL-PBR/tree/master/shaders. I adapted the code for use with gfx changed its style to reflect ours.

examples/aviator/main.rs
```diff
@@ -19,7 +19,7 @@ const COLOR_BROWN_DARK: three::Color = 0x23190f;
 const COLOR_BLUE: three::Color = 0x68c3c0;
 
 fn main() {
-    env_logger::init().unwrap();
+    env_logger::init();
```
vitvakatu: What about enabling logging for every example?

src/animation.rs
```diff
@@ -484,12 +489,14 @@ impl ActionData {
                         ))
                     };
                     let update = frame_start_value.slerp(frame_end_value, s);
+                    use Object;
```
vitvakatu: Let's import it globally.

examples/aviator/main.rs
kvark: would be nice, but requires some sort of example boilerplate, and creeps out of scope for this PR

src/animation.rs
kvark: oh god

.travis.yml
```diff
@@ -23,7 +23,8 @@ notifications:
     on_start: never
 
 script:
-  - cargo build
+  - cargo check
```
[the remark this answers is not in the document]
kvark: sure

src/material.rs
```diff
@@ -19,7 +19,7 @@ pub mod basic {
     /// Renders triangle meshes with a solid color or texture.
     #[derive(Clone, Hash, Debug, PartialEq, Eq)]
     pub struct Basic {
-        /// Solid color applied in the absense of `map`.
+        /// Solid color applied in the absence of `map`.
```
alteous: TIL.

kvark:
Thanks!
Bors r=vitvakatu

PR #185, opened 2018-02-24
Fix timed method in Delta implementation for axis::Key

### LINKED ISSUE #180: orbit controls don't update unless the mouse has moved
keeping the mouse steady while scrolling up and down does nothing.

---

os: windows 10
rust: latest nightly
three-rs: latest master

vitvakatu:
Aaaand again. Now the problem has been spotted in `gltf` example.

UPD: closes #180


vitvakatu:
Now also includes fix for mouse wheel

PR #189, opened 2018-02-25
Don't count keys hits while key is pressed

### LINKED ISSUE #187: `keys_hit()` does not properly report deltas
I put the following print statements inside my main event loop:

```rust
println!("{:?} {:?}",
    win.input.keys_hit().contains(&three::controls::Key::LWin), 
    win.input.keys_hit().contains(&three::controls::Key::Q));
```

If I press and hold the "LWin" button, I get only a single `true` printout. However, if I press and hold the "Q" button I get a `true` printout every iteration in the event loop. Is this expected behavior? What's the easiest way to get the `Delta` for "Q" presses?

I'm on MacOS.

vitvakatu:
closes #187 
The behavior of `sprite` and `lights` examples is changed slightly (now you can't hold the key), but I think now it's more logical

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -188,8 +188,10 @@ impl Input {
     ) {
         match state {
             ElementState::Pressed => {
-                self.state.keys_pressed.insert(key);
-                self.delta.keys_hit.push(key);
+                if !self.state.keys_pressed.contains(&key) {
+                    self.state.keys_pressed.insert(key);
+                    self.delta.keys_hit.push(key);
+                }
             }
             ElementState::Released => {
                 self.state.keys_pressed.remove(&key);


src/input/mod.rs
```diff
@@ -188,8 +188,10 @@ impl Input {
     ) {
         match state {
             ElementState::Pressed => {
-                self.state.keys_pressed.insert(key);
-                self.delta.keys_hit.push(key);
+                if !self.state.keys_pressed.contains(&key) {
```
kvark: let's use Entry API here: https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry

vitvakatu: It's not the case for entry API. There is no need in in-place modification of value and I need to check for value in hashmap to push key to the Vec.

kvark: the code accesses the `key` twice, so it *is* the case for entry API. Could be written as
```rust
if let Entry::Vacant(e) = self.state.keys_pressed.entry(key) {
  e.insert(()); // not sure what is this for a hash set
  self.delta.keys_hit.push(key);
}
```

vitvakatu: But there is no entry API for HashSet....

vitvakatu:
Oh, great. ICE on nightly.

vitvakatu:
bors r=kvark

PR #193, opened 2018-03-12
Instanced sprites entry point

### LINKED ISSUE #192: Sprite instancing
Sprite rendering has an important use case of tiled maps, e.g. https://github.com/gfx-rs/gfx/tree/pre-ll/examples/ubo_tilemap

We support instancing for meshes, and we should group all the tiles using the same texture (but different texel ranges) as well for instancing.

cc @eddyb @vitvakatu

kvark:
Fixes #192
TODO: instanced tile map example (in follow-ups)

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -304,12 +304,12 @@ pub(crate) struct GpuData {
     pub displacement_contributions: Vec<DisplacementContribution>,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 struct InstanceData {
-    pub slice: gfx::Slice<back::Resources>,
-    pub vertices: h::Buffer<back::Resources, Vertex>,
-    pub pso_data: PsoData,
-    pub material: Material,
+    slice: gfx::Slice<back::Resources>,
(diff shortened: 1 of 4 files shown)


src/render/mod.rs
```diff
@@ -945,15 +945,18 @@ impl Renderer {
...
+                        Some(ref map) => map.uv_range(),
+                        None => [0.0; 4],
+                    };
+                    if let Some(ref key) = gpu_data.instance_cache_key {
                         let vec = self.instance_cache.entry(key.clone()).or_insert((
                             InstanceData {
                                 slice: gpu_data.slice.clone(),
                                 vertices: gpu_data.vertices.clone(),
-                                pso_data: pso_data.clone(),
+                                pso_data: PsoData::Basic { color, map, param0 },
```
vitvakatu: I can't understand this change, but okay.

kvark: it's basically `pso_data` that has been destructured, so we compose it again
I'll add a comment

kvark:
@vitvakatu I ended up refactoring the instance data a bit

kvark:
bors r=vitvakatu

Issue #194, opened 2018-03-12
Fix build script


kvark:
Our current `build.rs` uses `includedir_codegen` with a wildcard for `rerun-if-changed`, which is not supported (also see https://github.com/tilpner/includedir/issues/3). Instead, we could manually iterate the shader files like WR does: https://github.com/servo/webrender/blob/master/webrender/build.rs
cc @eddyb

tilpner:
Upgrading to includedir 0.3.0 should fix this, thanks for reporting. :)

kvark:
Fixed by #193

PR #196, opened 2018-03-18
Try to improve compile time

vitvakatu:
It's extremely hard to say about some improvement on my machine (incremental compilation messes everything), @kvark, can you check it locally?


kvark:
The re-running criteria should be controlled by the `include_codegen` thing, see #194

PR #197, opened 2018-03-25
[WIP] Rework glTF loading API

randomPoison:
> NOTE: This PR is not yet ready to merge, but I would like to get feedback on the overall direction of this approach while I'm finishing the final details.

This PR proposes a major reworking of the API for loading and instantiating glTF assets. There are two primary goals for these changes:

* Expose more information from the source glTF file (e.g. the names of nodes, access to all scenes defined in the file, etc.).
* Make it possible to instantiate the contents of a glTF file multiple times without having to load the file from disk every time.

To accomplish this, I have split glTF loading into two parts:

1. Load the data and definitions from the glTF file into a `GltfDefinitions` object.
2. Use the `GltfDefinitions` to instantiate objects that can be added to the scene.

With this new API, loading a glTF files and adding its contents to a scene would go as follows:

```
// Load the definitions from the glTF file.
let definitions = window.factory.load_gltf(&path);

// Instantiate the first scene defined.
let instance = window.factory.instantiate_gltf_scene(&definitions, 0);
window.scene.add(&instance);
```

This only takes one more step than it did before, but it allows users to easily instantiate the contents of the file multiple times.

# Drawbacks

The current API isn't ideal from an ergonomics perspective. `GltfDefinitions` is setup to match the structure of the glTF file, which means objects reference each other by index. Even when you instantiate a scene, the nodes in a `GltfScene` are still setup to reference each other by index. I think this can be fixed using custom iterators and accessor types that abstract away the underlying indexing operations, similar to how the types in the `gltf` crate are implemented.

Currently I don't include node names in an instantiated `GltfScene`, which means you still need to use the `GltfDefinitions` to look up nodes by name. This could be improved by including names in `GltfScene`, at the cost of copying the names every time the user instantiates a scene.

Animations aren't instantiated automatically. To instantiate an animation, you have to specify a `GltfScene` that it should be applied to, and creating the animation can fail if it references a node that isn't present in the `GltfScene`. As far as I could tell, there was no information in the glTF spec that states which animations are part of which scenes, otherwise I would instantiate animations automatically.

---
This change is [Reviewable](https://reviewable.io/reviews/three-rs/three/197)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@
 # Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
 # More information here http://doc.crates.io/guide.html#cargotoml-vs-cargolock
 Cargo.lock
+
+# Configuration file for editors using the RLS.
+rls.toml
--- a/examples/gltf-morph-targets.rs
+++ b/examples/gltf-morph-targets.rs
@@ -12,12 +12,12 @@ fn main() {
     let default = concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/AnimatedMorphCube/AnimatedMorphCube.gltf");
     let path = std::env::args().nth(1).unwrap_or(default.into());
     let gltf = window.factory.load_gltf(&path);
-    window.scene.add(&gltf);
+    // window.scene.add(&gltf);
 
     let mut mixer = three::animation::Mixer::new();
-    for clip in gltf.clips {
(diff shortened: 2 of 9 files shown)


randomPoison:
After some discussion in chat, I'm going to take another approach to add the functionality I'm looking for. I'm going to close this PR in the meantime, as it could take a while to rework things.

PR #199, opened 2018-04-08
[WIP] Hierarchy templating API

randomPoison:
> NOTE: This PR is not yet ready to merge, but I would like to get feedback on the overall direction of this approach while I'm finishing the final details.

This is a follow-up to #197 that adds a more general-purpose system for instantiating copies of entire node hierarchies. The new API includes a `Hierarchy` type that contains all of the objects and describes the hierarchy of nodes, and a method `Factory::instantiate_hierarchy` to duplicate an existing hierarchy. `Factory::load_gltf` has been updated to return a `Vec<Hierarchy>`, with one `Hierarchy` for each scene described in the file. While `load_gltf` is currently the only way to easily construct a `Hierarchy`, the system is meant to be independent of glTF functionality and extra utilities could be added to make it easier for users to construct their own `Hierarchy` objects.

---
This change is [Reviewable](https://reviewable.io/reviews/three-rs/three/199)

### DIFF
--- a/src/factory/load_gltf.rs
+++ b/src/factory/load_gltf.rs
@@ -27,32 +27,54 @@ use geometry::{Geometry, Shape};
 use object::{self, Object};
 use super::Factory;
 
-/// The contents of a scene defined in a glTF file, fully insantiated and ready to be added to
-/// a scene and rendered.
+/// A glTF scene that has been instantiated and can be added to a [`Scene`].
 ///
-/// # Notes
+/// Created by instantiating a scene defined in a [`GltfDefinitions`] with
+/// [`Factory::instantiate_gltf_scene`]. A `GltfScene` can be added to a [`Scene`] with
+/// [`Scene::add`].
 ///
-/// The various contents
+/// # Examples
+///
+/// ```no_run
+/// # let mut window = three::Window::new("three-rs");
+/// let definitions = window.factory.load_gltf("my_data.gltf");
+///
+/// let gltf_scene = window.factory.instantiate_gltf_scene(&definitions, 0);
+/// window.scene.add(&gltf_scene);
+/// ```
+///
+/// [`Scene`]: struct.Scene.html
+/// [`Scene::add`]: struct.Scene.html#method.add
+/// [`GltfDefinitions`]: struct.GltfDefinitions.html
+/// [`Factory::instantiate_gltf_scene`]: struct.Factory.html#method.instantiate_gltf_scene
 #[derive(Debug, Clone)]
 pub struct GltfScene {
     /// A group containing all of the root nodes of the scene.
     ///
     /// While the glTF format allows scenes to have an arbitrary number of root nodes, all scene
     /// roots are added to a single root group to make it easier to manipulate the scene as a
-    /// whole. See `roots` for the full list of root nodes for the scene.
+    /// whole. See [`roots`] for the full list of root nodes for the scene.
+    ///
+    /// [`roots`]: #structfield.roots
     pub group: object::Group,
 
     /// The indices of the root nodes of the scene.
     ///
-    /// Each index corresponds to a node in `nodes`.
+    /// Each index corresponds to a node in [`nodes`].
+    ///
+    /// [`nodes`]: #structfield.nodes
     pub roots: Vec<usize>,
 
     /// The nodes that are part of the scene.
     ///
     /// Node instances are stored in a [`HashMap`] where the key is the node's index in the source
     /// [`GltfDefinitions::nodes`]. Note that a [`HashMap`] is used instead of a [`Vec`] because
-    /// the not all nodes in the source file are guaranteed to be used in the scene, and so node
+    /// not all nodes in the source file are guaranteed to be used in the scene, and so node
     /// indices in the scene instance may not be contiguous.
+    ///
+    /// [`HashMap`]: https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html
+    /// [`Vec`]: https://doc.rust-lang.org/stable/std/vec/struct.Vec.html
+    /// [`GltfDefinitions::nodes`]: struct.GltfDefinitions.html#structfield.nodes
     pub nodes: HashMap<usize, GltfNode>,
 }
 
@@ -156,9 +178,13 @@ pub struct GltfDefinitions {
 
 /// A template for a glTF mesh instance.
 ///
-/// Note that a glTF mesh doesn't map directly to three's concept of a [`Mesh`] (see
-/// [`GltfPrimitiveDefinition`] for a more direct analogy). Rather, `GltfMeshDefinition` can be instantated
-/// into a [`Group`] with one or more [`Mesh`] instances added to the group.
+/// Note that a glTF mesh doesn't map directly to three's [`Mesh`] type (see
+/// [`GltfPrimitiveDefinition`] for a more direct analogy). Rather, `GltfMeshDefinition` can
+/// be instantated into a [`Group`] with one or more [`Mesh`] instances added to the group.
+///
+/// [`Mesh`]: struct.Mesh.html
+/// [`GltfPrimitiveDefinition`]: struct.GltfPrimitiveDefinition.html
+/// [`Group`]: struct.Group.html
 #[derive(Debug, Clone)]
 pub struct GltfMeshDefinition {
     /// The name of the mesh template.
--- a/src/factory/load_gltf.rs
+++ b/src/factory/load_gltf.rs
@@ -460,6 +460,7 @@ fn instantiate_node_hierarchy<'a>(
 
     // Bind the skeleton to the node's meshes.
     if let Some(ref skeleton) = skeleton {
+        group.add(skeleton);
         for mesh in &mut meshes {
             mesh.set_skeleton(skeleton.clone());
         }
--- a/src/factory/load_gltf.rs
+++ b/src/factory/load_gltf.rs
@@ -102,11 +102,19 @@ pub struct GltfDefinitions {
     /// The scenes described in the glTF file.
     pub scenes: Vec<GltfSceneDefinition>,
 
+    /// The index of the default scene, if specified by the glTF file.
+    ///
+    /// The index corresponds to an element in `scenes`.
+    pub default_scene: Option<usize>,
+
     /// The skinned skeltons loaded from the glTF file.
     pub skins: Vec<GltfSkinDefinition>,
 
     /// The animation clips loaded from the glTF file.
     pub animations: Vec<GltfAnimationDefinition>,
+
+    /// Imported textures.
+    pub textures: Vec<Texture<[f32; 4]>>,
 }
 
 /// A template for a glTF mesh instance.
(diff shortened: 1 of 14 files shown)


kvark:
Thank you, amazing work! I really like what I'm seeing here, have a few concerns to be addressed before we proceed.

---

Reviewed 14 of 14 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/hub.rs, line 129 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9euW7Y4bRTrdGThbkN:-L9euW7ZAYT06ap-VwrT:b-79n631) ([raw file](https://github.com/three-rs/three/blob/4700abd3fb5b95abb39a0230007e19518bf1d133/src/hub.rs#L129)):*
> ```Rust
>     pub(crate) fn spawn(
>         &mut self,
>         sub: SubNode,
> ```

why did you have to make this change?

---

*[src/lib.rs, line 332 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9erkaVAY4WIGvUHZdx:-L9erkaVAY4WIGvUHZdy:b-z5b5mg) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/lib.rs#L332)):*
> ```Rust
> #[cfg(feature = "gltf-loader")]
> #[doc(inline)]
> pub use factory::load_gltf::*;
> ```

let's not have global imports here. Are there too many items?

---

*[src/lib.rs, line 366 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9erf1K2MZ0kSUXZy5c:-L9erf1L2od_UdTNi-g4:b-ywn2si) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/lib.rs#L366)):*
> ```Rust
> 
> #[doc(inline)]
> pub use template::{Hierarchy, HierarchyNode};
> ```

Did you consider naming it `Template` with `TemplateNode`?

---

*[src/node.rs, line 48 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9ersWzE_qUrVrgVRLE:-L9ersWzE_qUrVrgVRLF:bfh3mnk) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/node.rs#L48)):*
> ```Rust
>     pub(crate) fn shallow_clone(&self, sub: SubNode) -> NodeInternal {
>         NodeInternal {
>             visible: true,
> ```

why aren't we cloning the visibility flag?

---

*[src/template.rs, line 45 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9es3S008WuBiTIORII:-L9es3S008WuBiTIORIJ:b-tneq10) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/template.rs#L45)):*
> ```Rust
>     ///
>     /// [`nodes`]: #structfield.nodes
>     pub roots: Vec<usize>,
> ```

let's wrap the node index in an opaque new type

---

*[src/factory/load_gltf.rs, line 29 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9eslb1689ojYMWdRIh:-L9eslb1689ojYMWdRIi:bab73o3) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/factory/load_gltf.rs#L29)):*
> ```Rust
> use skeleton::Skeleton;
> use super::Factory;
> use template::*;
> ```

similarly, let's not do global imports here please

---

*[src/factory/mod.rs, line 1020 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9etz_r30c_fWRCIJQh:-L9etz_r30c_fWRCIJQi:bl5vmqy) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/factory/mod.rs#L1020)):*
> ```Rust
>     /// The new hierarchy will consist of new objects that duplicate the current state of
>     /// `hierarchy`.
>     pub fn instantiate_hierarchy(&mut self, hierarchy: &Hierarchy) -> Hierarchy {
> ```

This is not clear to me. I'd expect the user to have multiple instances of templates, but instantiating a template should give you a ready to use node (hierarchy/group), not another template

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/three-rs/three/199#-:-L9eua6v6-oCwqLJGPWy:bvwv8v6)*

randomPoison:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/hub.rs, line 129 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9euW7Y4bRTrdGThbkN:-L9fLMXV9AWgAKVXnj7W:b-uq1hi2) ([raw file](https://github.com/three-rs/three/blob/4700abd3fb5b95abb39a0230007e19518bf1d133/src/hub.rs#L129)):*

why did you have to make this change?
</blockquote></details>

This was to allow for spawning a fully-formed `NodeInternal` that already had a local transform supplied. For example, when instantiating a new `Group` we effectively want to clone the `NodeInternal` and then spawn that clone. The existing `spawn` only allowed for spawning a `SubNode` and would then create `NodeInternal` with a default transform, which isn't what I want in this case. I considered adding a new method such as `spawn_node`, but it seemed like it made more sense to make `spawn` more generally useful. Since `SubNode` can already be easily converted to `NodeInternal` with `into`, it was a minor change to update all the calling code to perform the conversion.

Now that I write this out, it occurs to me I could make the signature `spawn<T: Into<NodeInternal>>(..)` and then I wouldn't have to change any of the existing calls to `spawn`. If that'd be a preferred solution, let me know and I'll update it.

It's also worth noting that the current version of the code doesn't actually make use of this change currently. I'll likely go back to using this as I finish up the PR, though if I end up not needing it, then I'll revert the change.

---

*[src/lib.rs, line 332 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9erkaVAY4WIGvUHZdx:-L9fNj5t9X0kTPJEHpPC:beluopr) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/lib.rs#L332)):*

let's not have global imports here. Are there too many items?
</blockquote></details>

Nah, I did the glob import for convenience but I'll change it to manually list the imports.

---

*[src/lib.rs, line 366 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9erf1K2MZ0kSUXZy5c:-L9fNtLj1HTmB0bMMp4B:b-y5rhf1) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/lib.rs#L366)):*

Did you consider naming it `Template` with `TemplateNode`?
</blockquote></details>

Yeah, that's an option (and there's certainly room to bikeshed on naming). I'm a bit hesitant to go with `Template`, though, because "template" to me implies that it is only used to create objects, but can't be added to the scene directly, whereas `Hierarchy` acts like a normal object and can be added to the scene directly. I could name the module `hierarchy` instead, since there's currently an inconsistency there.

I don't feel too strongly about the naming, though, so I'm open to other suggestions or reasons why `Template` would be fine.

---

*[src/node.rs, line 48 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9ersWzE_qUrVrgVRLE:-L9fRavkDq5YVRK8d5Ws:b3l3zdw) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/node.rs#L48)):*

why aren't we cloning the visibility flag?
</blockquote></details>

I can't remember why I did that originally, but it seems like the wrong thing to do. I'll update that to preserve visibility.

---

*[src/template.rs, line 45 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9es3S008WuBiTIORII:-L9fSCB68mI3nY1As_fc:b-qg9hnq) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/template.rs#L45)):*

let's wrap the node index in an opaque new type
</blockquote></details>

Sure! I think that's a good starting point, but we'll probably want to do more work on the `Hierarchy` API to make it more fool-proof. For example, wrapping the indices in an opaque type would prevent users from using a random `usize` as a node ID, but they could still a node ID from one `Hierarchy` and use it in another, which would not be a valid thing to do.

Instead, I was thinking we'd be able to fully abstract away node IDs, and instead provide custom iterators that allow users to iterate over the contents of the hierarchy (and even add/remove nodes) without having to deal with node IDs directly. I was leaving that work for a follow-up PR, though, once the base API is working.

---

*[src/factory/load_gltf.rs, line 29 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9eslb1689ojYMWdRIh:-L9fWbOj8Zp1EyPKhlxN:bhy9ost) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/factory/load_gltf.rs#L29)):*

similarly, let's not do global imports here please
</blockquote></details>

Will do!

---

*[src/factory/mod.rs, line 1020 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9etz_r30c_fWRCIJQh:-L9fVGYU17vfwT0OVGu2:buk5hhd) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/factory/mod.rs#L1020)):*

This is not clear to me. I'd expect the user to have multiple instances of templates, but instantiating a template should give you a ready to use node (hierarchy/group), not another template
</blockquote></details>

`Hierarchy` is an `Object` and can be added to the scene directly. I figured that would be a good way to add an "offline" templating system without having to add too many new concepts to three. It also has the advantage that, when a function returns a `Hierarchy`, you don't have to instantiate it before adding it to the scene, e.g. this works if you only need one instance:

```
let gltf = window.factory.load_gltf("myModel.gltf");
window.scene.add(&gltf);
```

But if you want multiples, then this works too:

```
let gltf = window.factory.load_gltf("myModel.gltf");
for _ in 0..10 {
    let instance = window.factory.instantiate_hierarchy(&gltf);
    window.scene.add(&instance);
    instances.push(instance);
}
```

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/three-rs/three/199)*

kvark:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/hub.rs, line 129 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9euW7Y4bRTrdGThbkN:-L9jrMUG0NVv8d5ix2Nl:buxcufl) ([raw file](https://github.com/three-rs/three/blob/4700abd3fb5b95abb39a0230007e19518bf1d133/src/hub.rs#L129)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

This was to allow for spawning a fully-formed `NodeInternal` that already had a local transform supplied. For example, when instantiating a new `Group` we effectively want to clone the `NodeInternal` and then spawn that clone. The existing `spawn` only allowed for spawning a `SubNode` and would then create `NodeInternal` with a default transform, which isn't what I want in this case. I considered adding a new method such as `spawn_node`, but it seemed like it made more sense to make `spawn` more generally useful. Since `SubNode` can already be easily converted to `NodeInternal` with `into`, it was a minor change to update all the calling code to perform the conversion.

Now that I write this out, it occurs to me I could make the signature `spawn<T: Into<NodeInternal>>(..)` and then I wouldn't have to change any of the existing calls to `spawn`. If that'd be a preferred solution, let me know and I'll update it.

It's also worth noting that the current version of the code doesn't actually make use of this change currently. I'll likely go back to using this as I finish up the PR, though if I end up not needing it, then I'll revert the change.
</blockquote></details>

the generic solution you mentioned sounds good! if needed, obviously :)

---

*[src/lib.rs, line 366 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9erf1K2MZ0kSUXZy5c:-L9jrZYGARhanV2kxT2j:bahvxmu) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/lib.rs#L366)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

Yeah, that's an option (and there's certainly room to bikeshed on naming). I'm a bit hesitant to go with `Template`, though, because "template" to me implies that it is only used to create objects, but can't be added to the scene directly, whereas `Hierarchy` acts like a normal object and can be added to the scene directly. I could name the module `hierarchy` instead, since there's currently an inconsistency there.

I don't feel too strongly about the naming, though, so I'm open to other suggestions or reasons why `Template` would be fine.
</blockquote></details>

Well, I'm confused as to why we'd want `Hierarchy` to act like a normal object. Wouldn't it be cleaner to separate the concerns here?

---

*[src/template.rs, line 45 at r1](https://beta.reviewable.io/reviews/three-rs/three/199#-L9es3S008WuBiTIORII:-L9jrlg07Sa50CRbvpwH:b-maj4lo) ([raw file](https://github.com/three-rs/three/blob/91a04a59152c23a420302afdd128462f80266e0f/src/template.rs#L45)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

Sure! I think that's a good starting point, but we'll probably want to do more work on the `Hierarchy` API to make it more fool-proof. For example, wrapping the indices in an opaque type would prevent users from using a random `usize` as a node ID, but they could still a node ID from one `Hierarchy` and use it in another, which would not be a valid thing to do.

Instead, I was thinking we'd be able to fully abstract away node IDs, and instead provide custom iterators that allow users to iterate over the contents of the hierarchy (and even add/remove nodes) without having to deal with node IDs directly. I was leaving that work for a follow-up PR, though, once the base API is working.
</blockquote></details>

Sounds good!
If we do end up with IDs, we can always encode an extra private field in there carrying unique identification of the owner

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/three-rs/three/199)*

src/factory/load_gltf.rs
```diff
@@ -224,303 +182,357 @@ fn load_materials(
...
+    let children = node
+        .children()
+        .map(|child| child.index())
+        .collect();
+
+    // Attach the various objects to the group.
+    for mesh in &meshes { group.add(mesh); }
+    if let Some(ref camera) = camera { group.add(camera); }
+
+    // Recursively instantiate all children.
```
alteous: FYI, the previous version did not need to instantiate objects recursively.

randomPoison:
Closing this in favor of an alternative implementation, as discussed in chat.

Issue #125, opened 2017-10-02
Bake in default font


vitvakatu:
We should provide default font for text rendering as we provide default shaders. The [Karla](https://fonts.google.com/specimen/Karla) font is only 17 Kb.

alteous:
Perhaps we could leverage [`font_loader`](https://docs.rs/font-loader/0.4.2/font_loader/) also?

PR #200, opened 2018-04-16
Bake in default font

rukai:
Fixes https://github.com/three-rs/three/issues/125

I have used the Karla font as that was recommended in the issue.
Alternatively I could use the dejavu font already in the repo, which is not embedded anywhere, it is currently just loaded by an example.

This PR adds a function Factory::load_font_karla which returns a Font.
Alternatively I could change the Factory::ui_text to take an Option<Font> and then use the default font when its None.

Also adds an example to demonstrate how to use the text api.
Also fixes setting text color as that was setting alpha to 0, which was breaking my example.


src/factory/mod.rs
```diff
@@ -762,7 +762,13 @@ impl Factory {
...
                 file_path.display()
             ));
-        Font::new(buffer, file_path.to_owned(), self.backend.clone())
+        Font::new(buffer, format!("path: {:?}", file_path), self.backend.clone())
+    }
+
+    /// Load the Karla font
+    pub fn load_font_karla(&mut self) -> Font {
+        let buffer = include_bytes!("../../data/fonts/Karla-Regular.ttf").to_vec();
+        Font::new(buffer, String::from("Embeded Karla-Regular.ttf"), self.backend.clone())
```
kvark: somewhat minor, but we should totally be able to pass in a slice `&[u8]` in here instead of a vector

rukai: I tried just now but I run into lifetime errors I cant resolve.

vitvakatu: Shouldn't be possible due to internal organization of rusttype

kvark: Please elaborate. What's blocking the references here?

vitvakatu: Okay, our internal code. Static lifetime [here](https://github.com/three-rs/three/blob/master/src/text.rs#L90)

kvark: Does `GlyphBrush` inherit the lifetime from the raw data we pass in?

vitvakatu: yep, it does

vitvakatu: That's the reason I said it's `rusttype` deal mostly. I was forced to use `'static` and `Vec<u8>` while implementing this

src/text.rs
```diff
@@ -91,7 +91,7 @@ pub struct Font {
 }
 
 impl Font {
-    pub(crate) fn new(
+    pub(crate) fn new_owned(
```
kvark: can we re-use the [generic interface](https://docs.rs/gfx_glyph/0.10.1/gfx_glyph/fn.font.html) of `g::font` here to have a single signature accepting `T: Into<SharedBytes<'a>>`?

PR #203, opened 2018-05-02
Template System and Reworked glTF Handling

randomPoison:
This PR adds a new system for creating templates for hierarchies of objects. Templates describe many objects and their relationships to each other, and allow the user to instantiate the entire hierarchy of objects at once. Templates can be instantiated an arbitrary number of times. All instances of a template will use the same GPU resources for each instance of a mesh, reducing duplication of GPU resources and allowing for efficient instanced rendering.

* Adds a new module `template` with type declarations for `Template` and its various sub-objects.
* Adds `Factory::instantiate_template` for creating concrete objects from a template.
* Adds `Factory::instanced_geometry` to allow users to upload a `Geometry` to the GPU (this is needed to allow templates to reuse GPU resources).
* Removes `Factory::instantiate_mesh` and replace it with `Factory::instanced_mesh`, which allows the user to create a mesh given an `InstancedGeometry` and a `Material`. The removal of `instantiate_mesh` was suggested by @kvark, I opted to also add `instanced_mesh` as a convenience function for cases where constructing a full-blown `Template` is overkill.
* Reworks `Factory::load_gltf` to return a list of `Template` objects, returning one `Template` for each scene in the glTF file.
* Adds `Factory::camera` as a helper method to create a `Camera` from a `Projection`. Adding this simplified some of the logic in instantiating templates, since the template stores camera projections as `Projection`.
* Adds the option to set an up direction for the `Orbit` controller. I wound up not needing this for any of the examples, but I opted to leave it in the PR because it's useful functionality to have.

Sorry this took so long. <sub>Hopefully this attempt sticks. <sub>I'm gonna go to bed now.</sub></sub>

---
This change is [Reviewable](https://reviewable.io/reviews/three-rs/three/203)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@
 # Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
 # More information here http://doc.crates.io/guide.html#cargotoml-vs-cargolock
 Cargo.lock
+
+# Configuration file for editors using the RLS.
+rls.toml
--- a/examples/gltf-morph-targets.rs
+++ b/examples/gltf-morph-targets.rs
@@ -9,28 +9,33 @@ fn main() {
     window.scene.add(&light);
     window.scene.background = three::Background::Color(0xC6F0FF);
 
-    // Load the glTF file.
     let default = concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/AnimatedMorphCube/AnimatedMorphCube.gltf");
     let path = std::env::args().nth(1).unwrap_or(default.into());
+
+    // Load the contents of the glTF files. Scenes loaded from the file are returned as
+    // `Template` objects, which can be used to instantiate the actual objects for rendering.
     let template = window.factory.load_gltf(&path).pop().unwrap();
 
-    // Instantiate the contents of the file.
+    // Instantiate the contents of the template, and then add it to the scene.
     let (instance, animations) = window.factory.instantiate_template(&template);
     window.scene.add(&instance);
 
-    // Instantiate all of the animations in the glTF file and start playing them.
+    // Begin playing all the animations assoicated with the template we instantiated.
     let mut mixer = three::animation::Mixer::new();
     for animation in animations {
         mixer.action(animation);
     }
 
+    // Create a camera with which to render the scene, and control it with the built-in
+    // orbit controller, set to orbit the model.
     let camera = window.factory.perspective_camera(60.0, 0.1 .. 20.0);
-
     let mut controls = three::controls::Orbit::builder(&camera)
         .position([-3.0, 3.0, -3.0])
         .up([0.0, 1.0, 0.0])
         .build();
 
+    // Run the main loop, updating the camera controller, animations, and rendering the scene
+    // every frame.
     while window.update() && !window.input.hit(three::KEY_ESCAPE) {
         mixer.update(window.input.delta_time());
         controls.update(&window.input);
--- a/examples/gltf-morph-targets.rs
+++ b/examples/gltf-morph-targets.rs
@@ -9,22 +9,27 @@ fn main() {
     window.scene.add(&light);
     window.scene.background = three::Background::Color(0xC6F0FF);
 
+    // Load the glTF file.
     let default = concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/AnimatedMorphCube/AnimatedMorphCube.gltf");
     let path = std::env::args().nth(1).unwrap_or(default.into());
     let gltf = window.factory.load_gltf(&path);
-    window.scene.add(&gltf);
 
+    // Instantiate the contents of the file.
+    let instance = window.factory.instantiate_gltf_scene(&gltf, 0);
+    window.scene.add(&instance);
+
+    // Instantiate all of the animations in the glTF file and start playing them.
     let mut mixer = three::animation::Mixer::new();
-    for clip in gltf.clips {
+    for anim_def in &gltf.animations {
+        let clip = window.factory.instantiate_gltf_animation(&instance, anim_def).unwrap();
         mixer.action(clip);
     }
 
     let camera = window.factory.perspective_camera(60.0, 0.1 .. 20.0);
-    camera.set_position([0.0, 1.0, 5.0]);
 
     let mut controls = three::controls::Orbit::builder(&camera)
-        .position([-0.08, -0.05, 0.075])
-        .target([0.0, 0.0, 0.01])
+        .position([-3.0, 3.0, -3.0])
+        .up([0.0, 1.0, 0.0])
         .build();
 
     while window.update() && !window.input.hit(three::KEY_ESCAPE) {
--- a/examples/gltf-morph-targets.rs
+++ b/examples/gltf-morph-targets.rs
@@ -12,17 +12,16 @@ fn main() {
     // Load the glTF file.
     let default = concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/AnimatedMorphCube/AnimatedMorphCube.gltf");
     let path = std::env::args().nth(1).unwrap_or(default.into());
-    let gltf = window.factory.load_gltf(&path);
+    let template = window.factory.load_gltf(&path).pop().unwrap();
 
     // Instantiate the contents of the file.
-    let instance = window.factory.instantiate_gltf_scene(&gltf, 0);
+    let (instance, animations) = window.factory.instantiate_template(&template);
     window.scene.add(&instance);
 
     // Instantiate all of the animations in the glTF file and start playing them.
     let mut mixer = three::animation::Mixer::new();
-    for anim_def in &gltf.animations {
-        let clip = window.factory.instantiate_gltf_animation(&instance, anim_def).unwrap();
-        mixer.action(clip);
+    for animation in animations {
+        mixer.action(animation);
     }
 
     let camera = window.factory.perspective_camera(60.0, 0.1 .. 20.0);
--- a/examples/gltf-morph-targets.rs
+++ b/examples/gltf-morph-targets.rs
@@ -14,13 +14,13 @@ fn main() {
 
     // Load the contents of the glTF files. Scenes loaded from the file are returned as
     // `Template` objects, which can be used to instantiate the actual objects for rendering.
-    let template = window.factory.load_gltf(&path).pop().unwrap();
+    let templates = window.factory.load_gltf(&path);
 
     // Instantiate the contents of the template, and then add it to the scene.
-    let (instance, animations) = window.factory.instantiate_template(&template);
+    let (instance, animations) = window.factory.instantiate_template(&templates[0]);
     window.scene.add(&instance);
 
-    // Begin playing all the animations assoicated with the template we instantiated.
+    // Begin playing all the animations instantiated from the template.
     let mut mixer = three::animation::Mixer::new();
     for animation in animations {
         mixer.action(animation);
--- a/examples/gltf-node-animation.rs
+++ b/examples/gltf-node-animation.rs
@@ -9,12 +9,14 @@ fn main() {
     window.scene.add(&light);
     window.scene.background = three::Background::Color(0xC6F0FF);
 
-    // Load the glTF file.
     let default = concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/BoxAnimated/BoxAnimated.gltf");
     let path = std::env::args().nth(1).unwrap_or(default.into());
+
+    // Load the contents of the glTF files. Scenes loaded from the file are returned as
+    // `Template` objects, which can be used to instantiate the actual objects for rendering.
     let template = window.factory.load_gltf(&path).pop().unwrap();
 
-    // Instantiate the contents of the file.
+    // Instantiate the contents of the template, and then add it to the scene.
     let (instance, animations) = window.factory.instantiate_template(&template);
     window.scene.add(&instance);
 
@@ -24,14 +26,16 @@ fn main() {
         mixer.action(animation);
     }
 
+    // Create a camera with which to render the scene, and control it with the built-in
+    // orbit controller, set to orbit the model.
     let camera = window.factory.perspective_camera(60.0, 0.1 .. 100.0);
-    window.scene.add(&camera);
-
     let mut controls = three::controls::Orbit::builder(&camera)
         .position([3.0, 3.0, 3.0])
         .target([0.0, 1.0, 0.0])
         .build();
 
+    // Run the main loop, updating the camera controller, animations, and rendering the scene
+    // every frame.
     while window.update() && !window.input.hit(three::KEY_ESCAPE) {
         mixer.update(window.input.delta_time());
         controls.update(&window.input);
--- a/examples/gltf-node-animation.rs
+++ b/examples/gltf-node-animation.rs
@@ -14,13 +14,13 @@ fn main() {
 
     // Load the contents of the glTF files. Scenes loaded from the file are returned as
     // `Template` objects, which can be used to instantiate the actual objects for rendering.
-    let template = window.factory.load_gltf(&path).pop().unwrap();
+    let templates = window.factory.load_gltf(&path);
 
     // Instantiate the contents of the template, and then add it to the scene.
-    let (instance, animations) = window.factory.instantiate_template(&template);
+    let (instance, animations) = window.factory.instantiate_template(&templates[0]);
     window.scene.add(&instance);
 
-    // Instantiate all of the animations in the glTF file and start playing them.
+    // Start playing all the animations from the template.
     let mut mixer = three::animation::Mixer::new();
     for animation in animations {
         mixer.action(animation);
--- a/examples/gltf-node-animation.rs
+++ b/examples/gltf-node-animation.rs
@@ -9,13 +9,19 @@ fn main() {
     window.scene.add(&light);
     window.scene.background = three::Background::Color(0xC6F0FF);
 
+    // Load the glTF file.
     let default = concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/BoxAnimated/BoxAnimated.gltf");
     let path = std::env::args().nth(1).unwrap_or(default.into());
     let gltf = window.factory.load_gltf(&path);
-    window.scene.add(&gltf);
 
+    // Instantiate the contents of the file.
+    let instance = window.factory.instantiate_gltf_scene(&gltf, 0);
+    window.scene.add(&instance);
+
+    // Instantiate all of the animations in the glTF file and start playing them.
     let mut mixer = three::animation::Mixer::new();
-    for clip in gltf.clips {
+    for anim_def in &gltf.animations {
+        let clip = window.factory.instantiate_gltf_animation(&instance, anim_def).unwrap();
         mixer.action(clip);
     }
 
--- a/examples/gltf-node-animation.rs
+++ b/examples/gltf-node-animation.rs
@@ -12,17 +12,16 @@ fn main() {
     // Load the glTF file.
     let default = concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/BoxAnimated/BoxAnimated.gltf");
     let path = std::env::args().nth(1).unwrap_or(default.into());
-    let gltf = window.factory.load_gltf(&path);
+    let template = window.factory.load_gltf(&path).pop().unwrap();
 
     // Instantiate the contents of the file.
-    let instance = window.factory.instantiate_gltf_scene(&gltf, 0);
+    let (instance, animations) = window.factory.instantiate_template(&template);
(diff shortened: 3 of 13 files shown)


kvark:
Amazing work! I'm happy to see this moving forward :+1: 
Left a few concerns/questions for discussion here.

---

Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[examples/gltf-pbr-shader.rs, line 35 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWFNDC4wfLPCrKW6rd:-LBWFNDC4wfLPCrKW6re:b-mzz9eq) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/examples/gltf-pbr-shader.rs#L35)):*
> ```Rust
> 
>     // If we didn't find a camera in the glTF scene, create a default one to use.
>     let cam = cam.unwrap_or_else(|| {
> ```

does that look ok when running the example with no parameters?

---

*[src/template.rs, line 4 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWLFxy016SJVY8rt40:-LBWLFxy016SJVY8rt41:b-y4ii6u) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L4)):*
> ```Rust
> //!
> //! It is often the case that you will want to have multiple instances of the same model or
> //! hierarchy of object in your scene. While you could manually construct each instance yourself,
> ```

nit: hierarchy of objects
nit: rewrite to refer to "user" as opposed to "you"

---

*[src/template.rs, line 5 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWLPIF-tQP730Ln3ti:-LBWLPIF-tQP730Ln3tj:b-ddwij) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L5)):*
> ```Rust
> //! It is often the case that you will want to have multiple instances of the same model or
> //! hierarchy of object in your scene. While you could manually construct each instance yourself,
> //! three provides a templating system to allow you to describe your model's hierarchy ahead
> ```

make "three" distinguishable here, by either appending the -rs prefix, or applying some formatting

---

*[src/template.rs, line 22 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWMLsK3CWPk-q45y9m:-LBWMLsK3CWPk-q45y9n:bfzotas) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L22)):*
> ```Rust
> //! information on the different node sub-types.
> //!
> //! Data for specific types of nodes are held in the other member arrays of [`Template`]. For
> ```

big thanks for writing the docs up front!

---

*[src/template.rs, line 47 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWN3f48ogRiYw7ar1G:-LBWN3f48ogRiYw7ar1H:b2vg9op) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L47)):*
> ```Rust
> //!
> //! When setting up a mesh in a template, you must first upload your [`Geometry`] to the GPU
> //! using [`Factory::instanced_geometry`]. This will give you an [`InstancedGeometry`] object
> ```

should be called `instantiate_geometry` since it's an expensive operation and not a simple accessor

---

*[src/template.rs, line 114 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWNgqVDfK0ZqCAv-ci:-LBWNgqW4_Sdu-qbf5te:b-ah1qz6) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L114)):*
> ```Rust
> 
>     /// All objects defined by this template.
>     pub nodes: Vec<TemplateNode>,
> ```

perhaps, we could me more consistent in naming here and call it `NodeTemplate`?

---

*[src/template.rs, line 158 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWO5yI7KVZDeo9BUHq:-LBWO5yJEP5y-Z0zjDsJ:b-xnyljq) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L158)):*
> ```Rust
>     ///
>     /// This translation is relative to its parent node when instantiated.
>     pub translation: mint::Point3<f32>,
> ```

perhaps, we should re-use `node::Transform` here

---

*[src/template.rs, line 210 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWOzH524awJ1n2NQGG:-LBWOzH6CQMnbddWmiFL:bh5kcqz) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L210)):*
> ```Rust
> /// [module documentation]: ./index.html#node-templates
> #[derive(Debug, Clone)]
> pub enum TemplateNodeData {
> ```

Note that we can technically simplify the structure complexity here if we reverse the relationships: have all the specific templates to just have `node: usize` field pointing to a node template, which becomes just a transform.
Doing this would make some things simpler, but we need to ensure we'll still be able to instantiate the hierarchy into our internal node tree

---

*[src/factory/mod.rs, line 574 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWQn3_3s__ahFnn9KR:-LBWQn3a8YkpNjmpMA2f:bokd10h) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/factory/mod.rs#L574)):*
> ```Rust
>     ///
>     /// [`template`]: ./template/index.html#mesh-instancing
>     pub fn instanced_geometry(
> ```

we need to hide this step for anything but advanced use of the engine

---

*[src/factory/mod.rs, line 633 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWQv5J4cqz--UMQXk9:-LBWQv5J4cqz--UMQXkA:b1p8oae) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/factory/mod.rs#L633)):*
> ```Rust
>     /// [`Mesh`]: ./struct.Mesh.html
>     /// [`template`]: ./template/index.html#mesh-instancing
>     pub fn instanced_mesh<M: Into<Material>>(
> ```

similarly, this should be called with a verb ("instantiate_")

---


*Comments from [Reviewable](https://reviewable.io/reviews/three-rs/three/203#-:-LBWR0fU6UKWMmvzL4sY:b-54atoc)*

kvark:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/template.rs, line 47 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWN3f48ogRiYw7ar1G:-LBXK6ymABdGkDsO2OAR:bquzjic) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L47)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

For `instanced_geometry` and `instanced_mesh` I followed the naming convention for creating other objects, e.g. `Factory::group`, `Factory::mesh`, and such. I can change them to use `instantiate_` instead, but I'm not sure that makes the most sense. Is there a reason why we would use a different naming convention for these methods? 
</blockquote></details>

we should probably rename everything to align better with rust conventions, e.g. create_mesh, etc

---

*[src/template.rs, line 210 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWOzH524awJ1n2NQGG:-LBXKMqk2Hrw0RD-Xsvc:b-yxjitv) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L210)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

I'm not sure I'm following what you're suggesting. Could you clarify further, maybe provide a rough sketch of what you're imagining?
</blockquote></details>

What I'm saying is that we can get away without the `TemplateNodeData` enum. For example, instead of having `Mesh(usize)` variant here, there could be a `MeshTemplate::node` field with the node index. Doing that for other XXXTemplate structs would eventually lead to removal of all variants of this enum, which means we can get away without it entirely.
Is this a little more clear?

---

*[src/factory/mod.rs, line 574 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWQn3_3s__ahFnn9KR:-LBXL8z-0XTyYqTzsUXh:brlawkd) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/factory/mod.rs#L574)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

Agreed that this only necessary for advanced usage. Is there anything I should do to "hide" it for less advanced users?
</blockquote></details>

not really

---


*Comments from [Reviewable](https://reviewable.io/reviews/three-rs/three/203)*

randomPoison:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/template.rs, line 4 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWLFxy016SJVY8rt40:-LBYXM1SCGLhUk0MiCmF:bd8fpii) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L4)):*

nit: hierarchy of objects
nit: rewrite to refer to "user" as opposed to "you"
</blockquote></details>

Is there an established convention for not using second person in Rust docs? I did a quick look through the standard library docs, and it seems like second person is used pretty commonly, especially in module-level docs (e.g. [`std` crate docs](https://doc.rust-lang.org/stable/std/index.html) and [`collections` module docs](https://doc.rust-lang.org/stable/std/collections/index.html)).

---

*[src/template.rs, line 5 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWLPIF-tQP730Ln3ti:-LBYY1Dw9DABL9uHelI1:b-896fix) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L5)):*

make "three" distinguishable here, by either appending the -rs prefix, or applying some formatting
</blockquote></details>

Done.

---

*[src/template.rs, line 22 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWMLsK3CWPk-q45y9m:-LBYY3uR1mWWDGa43_Lc:b-kdol31) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L22)):*

big thanks for writing the docs up front!
</blockquote></details>

No problem! <sub>It was only the most time consuming portion of the entire PR...</sub>

---

*[src/template.rs, line 47 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWN3f48ogRiYw7ar1G:-LBYYSUQ8EJhzQucgUmh:bhvqf0p) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L47)):*

we should probably rename everything to align better with rust conventions, e.g. create_mesh, etc
</blockquote></details>

Ah, that's reasonable. I updated them, but I don't think it would make sense to change the other methods in this PR.

For the two methods in this PR, I went with `upload_geometry` (to match the language I use in the documentation) and `create_instanced_mesh`.

---

*[src/template.rs, line 114 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWNgqVDfK0ZqCAv-ci:-LBY_4QU8rZZqb2Ozm-9:b-896fix) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L114)):*

perhaps, we could me more consistent in naming here and call it `NodeTemplate`?
</blockquote></details>

Done.

---

*[src/template.rs, line 158 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWO5yI7KVZDeo9BUHq:-LBYdjTF4z8H_75ZXGS2:b-896fix) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L158)):*

perhaps, we should re-use `node::Transform` here
</blockquote></details>

Done.

---

*[src/template.rs, line 210 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWOzH524awJ1n2NQGG:-LBYe4RoEqYcyMOWgNVl:b-rhntrx) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L210)):*

What I'm saying is that we can get away without the `TemplateNodeData` enum. For example, instead of having `Mesh(usize)` variant here, there could be a `MeshTemplate::node` field with the node index. Doing that for other XXXTemplate structs would eventually lead to removal of all variants of this enum, which means we can get away without it entirely.
Is this a little more clear?
</blockquote></details>

So, if I'm understanding correctly, you're suggesting something like this:

```
pub struct Template {
    pub nodes: Template<Transform>,
    pub groups: Template<GroupTemplate>,

    // Members for each type of object.
}

pub struct MeshTemplate {
    pub node: usize,

    // The same members as before...
}

pub struct GroupTemplate {
    pub node: usize,

    pub children: Vec<???>,
}
```

That makes sense to me. My only question at this point is how do group templates specify their children?

```
pub children: Vec<???>,
```

They can't just specify children as a `Vec<usize>` because then there's no way to tell what type of object each child refers to. There are two options that I can think of:

* Create an enum `ChildType`. This would have effectively the same variants that`TemplateNodeData` has, but they would at least be simplified to all wrapping a `usize`.
* Have a separate list of children for each type of object (e.g. `mesh_children`, `group_children`, etc.). This option feels bad to me; It makes it harder to iterate over all children of a group, and it involves multiple allocations where one would do.

Do you have a preference as to what approach to take? I'd lean towards the enum, personally. Or is there another approach that I'm not seeing?

---

*[src/factory/mod.rs, line 633 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWQv5J4cqz--UMQXk9:-LBYdl08EfOgjMcYEe-x:b-896fix) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/factory/mod.rs#L633)):*

similarly, this should be called with a verb ("instantiate_")
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/three-rs/three/203)*

kvark:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/template.rs, line 4 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWLFxy016SJVY8rt40:-LBbRMoQ85BjGCmkH2rs:b-3eces0) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L4)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

Is there an established convention for not using second person in Rust docs? I did a quick look through the standard library docs, and it seems like second person is used pretty commonly, especially in module-level docs (e.g. [`std` crate docs](https://doc.rust-lang.org/stable/std/index.html) and [`collections` module docs](https://doc.rust-lang.org/stable/std/collections/index.html)).
</blockquote></details>

Interesting findings! I'll withdraw my nit then, for now at least :)

---

*[src/template.rs, line 47 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWN3f48ogRiYw7ar1G:-LBbRXhtEmP2OdMl3h2Y:bq0ncok) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L47)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

Ah, that's reasonable. I updated them, but I don't think it would make sense to change the other methods in this PR.

For the two methods in this PR, I went with `upload_geometry` (to match the language I use in the documentation) and `create_instanced_mesh`.
</blockquote></details>

sounds good

---

*[src/template.rs, line 210 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWOzH524awJ1n2NQGG:-LBbdJYp6Bu5T2AmLjmH:b-9w0jl4) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L210)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

So, if I'm understanding correctly, you're suggesting something like this:

```
pub struct Template {
    pub nodes: Template<Transform>,
    pub groups: Template<GroupTemplate>,

    // Members for each type of object.
}

pub struct MeshTemplate {
    pub node: usize,

    // The same members as before...
}

pub struct GroupTemplate {
    pub node: usize,

    pub children: Vec<???>,
}
```

That makes sense to me. My only question at this point is how do group templates specify their children?

```
pub children: Vec<???>,
```

They can't just specify children as a `Vec<usize>` because then there's no way to tell what type of object each child refers to. There are two options that I can think of:

* Create an enum `ChildType`. This would have effectively the same variants that`TemplateNodeData` has, but they would at least be simplified to all wrapping a `usize`.
* Have a separate list of children for each type of object (e.g. `mesh_children`, `group_children`, etc.). This option feels bad to me; It makes it harder to iterate over all children of a group, and it involves multiple allocations where one would do.

Do you have a preference as to what approach to take? I'd lean towards the enum, personally. Or is there another approach that I'm not seeing?
</blockquote></details>

Yeah, in this approach the group would not specify it's children (so, neither of the points in your list). This would be implicit, based on which objects point to the node associated with the group. I'm not 100% sure this is feasible, but it does feel important to explore here.

---

*[src/factory/mod.rs, line 2 at r2](https://reviewable.io/reviews/three-rs/three/203#-LBbd1bm5RW1u8X5zM1_:-LBbd1bm5RW1u8X5zM1a:bgo40kz) ([raw file](https://github.com/three-rs/three/blob/ae201a19d10e22ab98af9b0ccbb0d6882f3d80a9/src/factory/mod.rs#L2)):*
> ```Rust
> #[cfg(feature = "gltf-loader")]
> pub(crate) mod load_gltf;
> ```

why does it need to be `pub`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/three-rs/three/203)*

kvark:
Reviewed 4 of 4 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/template.rs, line 210 at r1](https://reviewable.io/reviews/three-rs/three/203#-LBWOzH524awJ1n2NQGG:-LBg318ZA2mFUvkPgZda:b-ml3hg0) ([raw file](https://github.com/three-rs/three/blob/d77d90d2622c1cf94eb32d8d003ec8e0aa00658b/src/template.rs#L210)):*
<details><summary><i>Previously, randomPoison (David LeGare) wrote…</i></summary><blockquote>

Totally, I'm down to play in the (design) space. Let me give a concrete example to see if I'm understanding.

Let's say I want to create a template for a Group at position (1, 2, 3) and give it two children, a mesh and a light. To describe this I would create a `GroupTemplate` and have it's `node` member point to a `NodeTemplate` with a position at (1, 2, 3). Then, I would make a `MeshTemplate` and `LightTemplate`, and have each of their `node` members point to the same `NodeTemplate` that the `GroupTemplate` points to, which would implicitly mark them as being children of the group. Is that right? That's the only setup I can think of that would implicitly associate objects with a group based on their `node`.

Assuming that's right, how would I specify a different transform for the mesh or the node (e.g. I want the mesh to have a local position of (4, 5, 6))? As I understand it, `GroupTemplate` is the only one that would be able to specify it's own transform via `node`, for other types `node` would effectively specify their parent, and they would end up with a default local transform.

An alternative that occurs to me now: Instead of having groups list their children, have all the types have a `group: Option<usize>` which specifies the index of the group they should be added to. That would achieve your goal of being able to remove `TemplateNodeData`, while keeping parent/child relationships explicit (which I kind of prefer). I generally prefer to have parents list their children, rather than the other way around, but this approach would help to simplify the overall structure of the template, and has the nice benefit of reducing the number of allocations needed. It might also simplify the template instantiation code, now that I think about it 🤔
</blockquote></details>

Sounds like you are on the right track :) let's follow up on Gitter with further concerns and findings

---


*Comments from [Reviewable](https://reviewable.io/reviews/three-rs/three/203)*

randomPoison:
Sure! I'll squash out all the unimportant commits.

kvark:
:rocket: :sheep: it
bors r+

PR #210, opened 2018-10-05
Update rodio to 0.8

est31:


est31:
Note that this is untested because I couldn't find any audio/sound examples. But cargo test still works and it still compiles.

vitvakatu:
Thank you.
I was pretty sure we have example with audio (we even have `music.flac` in test_data!). Perhaps we can add one.

est31:
Sure, if you add an example, I'd love to test it.

PR #215, opened 2019-01-01
Add the ability to set the full screen mode after creating the window.

alexander-irbis:


src/window.rs
```diff
@@ -102,6 +103,7 @@ impl Builder {
         } else {
             None
         };
+        let is_fullscreen = monitor_id.is_some();
```
kvark: nit: just use `self.fullscreen` instead

kvark:
Can you squash, please?

PR #217, opened 2019-01-01
Add the ability to change light settings.

alexander-irbis:


src/light.rs
```diff
@@ -44,6 +50,24 @@ impl Ambient {
     pub(crate) fn new(object: Base) -> Self {
         Ambient { object }
     }
+
+    /// Change light color.
+    pub fn set_color(
```
kvark: sounds like we could have a `Light` trait with those methods? Maybe even have the implementation shared somehow

alexander-irbis: I have some concerns. 
We can use the trait and implement methods in it, then every time we want to use these methods, we should import the trait, it's annoying.
On the other hand, the implemntation with a macro is unfriendly for IDEs.
Do you prefer the case with a trait?

kvark:
Agreed, there is a bit of an annoyance with a trait. Let's proceed.
Bors r+

PR #225, opened 2019-04-23
Make audio support optional

linkmauve:
This reduces the amount of dependencies of 16 if audio support isn’t required by the game.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,9 @@ exclude = ["doc/*", "bors.toml", ".travis.yml", "test_data/*"]
 [lib]
 
 [features]
-default = ["opengl"]
+default = ["opengl", "audio"]
 opengl = ["gfx_device_gl", "gfx_window_glutin", "glutin"]
+audio = ["rodio"]
 
 [build-dependencies]
 includedir_codegen = "0.5"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -15,8 +15,9 @@ exclude = ["doc/*", "bors.toml", ".travis.yml", "test_data/*"]
 [lib]
 
 [features]
-default = ["opengl"]
+default = ["opengl", "audio"]
 opengl = ["gfx_device_gl", "gfx_window_glutin", "glutin"]
+audio = ["rodio"]
```
kvark: would you consider changing the feature name to simply "rodio"?

linkmauve: Sure, but that sounds less like something the user may understand, at first sight.

kvark: yeah, I take the point. Let's proceed then

kvark:
It just occurred to me that we are missing CI checks for builds without "audio" feature

linkmauve:
This might work, let’s wait for CI to run.

linkmauve:
Turns out three doesn’t currently build when the opengl feature is absent, I force-enabled it in this commit. It would be useful to fix that eventually.

linkmauve:
Seems to test just fine, thanks for the comment!

kvark:
Amazing work, thank you!
bors r+

PR #227, opened 2019-06-26
Gui

cedric-h:
I tried to run `cargo fmt`, but that didn't seem to work at all.
I think that's why almost every file was changed. I saw your .rustfmt, so I figured it was configured the way you guys wanted it.
The files you should be looking at, then, are src/window.rs, src/gui.rs, and examples/gui.rs.
I broke backwards compatibility because of the Window struct, right now for all of the other examples you'd have to do `Window<gui::NoBackend>::new()` to make a new window.
I'm not sure how you'd want me to fix that, whether you'd want me to `cfg` out everything GUI related, or maybe make what is now the window into `GuiWindow`, and then typedef `Window` as `GuiWindow<NoBackend>`. Either of these and several other things could preserve backwards compatibility, but it's up to you guys.


cedric-h:
Naturally my PR doesn't pass the tests, but that's because of the window instantiation backwards compat thing, like I explained above.

ecton:
@kvark Hey, I'm loving this library and was looking into updating the Cargo.toml to get onto current versions of libraries, and in the process of fixing some of that, the rustfmt.toml issue that I found on gitter came up as I edited factory/mod.rb. In current versions of toolchains, this file shrinks drastically because max_width now appears to transform all multiline expressions, such as the one you mentioned in the last comment on here.

What is your current opinion on how the formatting of files for this project should happen? I'd suggest removing the max_width directive, and using `#[rustfmt::skip]` in places where you want to control the formatting specifically -- I'm guessing mostly the data arrays?

Sorry for hijacking this PR for the question -- it seemed relevant to this issue's conversation. Great library, btw!

kvark:
@ecton this is indeed unrelated to the PR:) Please open an issue about rust formatting (or a PR with the suggest changes). Roughly, your proposal looks reasonable to me.

Issue #222, opened 2019-03-05
Port to wgpu-rs


kvark:
This is a conflicting proposal with #175

Pros:
  - dog-fooding `wgpu`, while still relying on gfx-hal
  - all the platforms for free
  - run on the Web

Cons: NONE

vitvakatu:
@kvark any progress on this? If not, I might give it a try, possibly in several iterations.

kvark:
@vitvakatu glad to see you here!

I haven't started that port, been very busy with gfx-rs/portability, wgpu-rs, and vange-rs. For the near future, I'll be occupied with work as well as trying to integrate wgpu-rs into Gecko, so I'll not have time to do any three-rs work. Speaking of which, I'm sure it's it would be lots of run to port!

However, I can think of one thing that would be more important for the project at this stage than porting to wgpu-rs. This is - evaluation of how well the current approach worked out, a sort of mid-mortem. Obviously, it's not ideal: three.js is a JS API that doesn't always map nicely to Rust, the hidden node tree wasn't obvious for users to consider (it doesn't map to Rust owning/borrowing mechanic), and overall the model seems quite restricted in what you can do without modifying three-rs itself.

So, if you have some bandwidth to work on it, try to evaluate if three-rs turned out to be the right API to begin with. If a different project needs to be born on the ashes of it, instead, and what it would look like. Or not, and we just need to port it to wgpu-rs and simply get more cool stuff built on it!

vitvakatu:
Seems reasonable. A long time ago I tried to use Three as an engine for pretty simple [Asteroids game](https://github.com/vitvakatu/asteroids) - and some things were hard to make using current API.

So I need to look into our current API and also learn new ideas in Rust gamedev ecosystem to make a conclusion.

Nazariglez:
> * r

Just a question, wgpu supports web? In the readme of the wgpu says that don't support it.

kvark:
It doesn't target the web yet. Once gfx-rs WebGL backend becomes available, we'll enable it here. Another path will open once the browsers start supporting WebGPU.

Nazariglez:
I'm working on a Cross Platform Layer (Audio, Inputs, Graphics) to use as a foundation for apps or games, and I'm looking to wgpu.rs as an alternative to webgl/opengl. Maybe in a few months. Thanks!

dreammaker:
Hi.  First of all, thanks for creating this library!  It's pretty awesome that it exists!

[...]

Issue #83, opened 2017-08-24
Web Assembly Support


Schr3da:
Is web assembly initially planned?

kvark:
@Schr3da we haven't look into it, and don't know what the requirements/implications are.
It would be much welcome for sure ;)

jpryne:
AFAICT, WASM has the same limitations as WebGL. If Three.rs can run using OpenGL ES 2.0, then it should compile to WASM just fine. I haven't tested this yet, but am digging into Three.rs for just this reason...

alteous:
I imagine there will only be a few issues to consider (all amendable) before WASM support can be a reality.

1. Faces take `GL_UNSIGNED_INT` indices. We would have to enable the [`OES_element_index_uint`](https://developer.mozilla.org/en-US/docs/Web/API/OES_element_index_uint) extension.
2. Shaders are written in more recent GLSL versions. We might have to write equivalent GLSL 1.x shaders.
3. Use of Rust `mpsc` channels. This might have to be replaced with a WASM-friendly equivalent.
4. Compilation, although I've had success with `cargo web`.

kvark:
Note that gfx-0.17 has WebGL2 support, so we may be very close to get this going.
I'm planning on updating our graphics stack, and I'll test WebGL2 for three-rs on the way, will report back here.

ForsakenHarmony:
is there any progress / timeline for this?

kvark:
@ForsakenHarmony we are currently stopped in a situation where Emscripten could possibly be gotten to work. The upstream PR unblocking it has landed only recently - https://github.com/kripken/emscripten/pull/5951 . However, Rust community/ecosystem appears to be moving in the direction of using `wasm-unknown-unknown` target instead (thus, no Emscripten), and we are watching a hero laying down this path in  https://github.com/gfx-rs/gfx/issues/1900

Anyhow, a champion is always needed here in Three-rs to get things hooked up and tested. Jump on our [gitter](https://gitter.im/three-rs/Lobby) to chat about it if you want to lend a hand ;)

ForsakenHarmony:
I'm not sure if I could be of any use

LordBenjamin:
Any thoughts on this now that gfx-rs supports `wasm-unknown-unknown`?

I'm interested and could possibly lend some time to implementing (although I'm a Rust newbie) if there is a good idea of what is required?

jpryne:
[...]

PR #230, opened 2019-09-09
update to latest glutin

happydpc:


src/render/mod.rs
```diff
@@ -538,13 +539,14 @@ impl Renderer {
...
+        context: glutin::ContextBuilder<NotCurrent>,
         event_loop: &glutin::EventsLoop,
         source: &source::Set,
-    ) -> (Self, glutin::GlWindow, Factory) {
+    ) -> (Self, glutin::WindowedContext<PossiblyCurrent>, Factory) {
         use gfx::texture as t;
 
-        let (window, device, mut gl_factory, out_color, out_depth) = gfx_window_glutin::init(builder, context, event_loop).unwrap();
+        let (_windowedContext, device, mut gl_factory, out_color, out_depth) = gfx_window_glutin::init(builder, context, event_loop).unwrap();
+        let window = _windowedContext.window();
```
vitvakatu: Please don't use underscore in front of regular variables. `windowedContext` doesn't need it

src/window.rs
```diff
@@ -111,6 +112,8 @@ impl Builder {
             .with_title(self.title.clone());
 
         let context = glutin::ContextBuilder::new()
+            .with_gl_profile(GlProfile::Core)
+            .with_gl(GlRequest::Latest)
```
vitvakatu: I'd rather prefer to see a specific version here. @kvark what do you think?

src/window.rs
```diff
@@ -147,12 +150,13 @@ impl Builder {
             try_override!(basic, gouraud, pbr, phong, quad, shadow, skybox, sprite,);
         }
 
-        let (renderer, window, mut factory) = Renderer::new(builder, context, &event_loop, &source_set);
-        let dpi = window.get_hidpi_factor();
+        let (renderer, windowedContext, mut factory) = Renderer::new(builder, context, &event_loop, &source_set);
+        let dpi = windowedContext.window().get_hidpi_factor();
         let scene = factory.scene();
+//        let window = windowedContext.window();
```
vitvakatu: Please remove commented code

src/window.rs
```diff
@@ -191,16 +195,17 @@ impl Window {
             input.reset();
         }
 
-        self.window.swap_buffers().unwrap();
-        let window = &self.window;
+//        self.windowedContext.window().swap_buffers().unwrap();
```
vitvakatu: Here as well

src/window.rs
```diff
@@ -242,17 +247,18 @@ impl Window {
...
+        let size = self.windowedContext
+            .window()
             .get_inner_size()
             .expect("Can't get window size")
             .to_physical(self.dpi);
         [size.width as f32, size.height as f32].into()
     }
 
-    /// Returns underlaying `glutin::GlWindow`.
+    /// Returns underlaying `glutin::Window`.
```
vitvakatu:
```suggestion
    /// Returns underlaying `glutin::WindowedContext`.
```

src/window.rs
kvark: Some older GPUs only support GL 2.1, while some macs do *not* support older GL 2x. So if we are completely pedantic, we can't have a specific version requirement (even though, practically, GL 3.2 would fit most cases). So having `Latest` here is fine.

happydpc:
Sorry for these small problems. Will three.rs plan to support gfx-hal, because I found that the opengl was not that enough. How about supporting rendy ?

kvark:
@happydpc I think it would make more sense to base on wgpu-rs for this project instead of gfx/rendy. See https://github.com/three-rs/three/issues/83#issuecomment-529161223 and #222

happydpc:
Yes, that's true. I had a try on wgpu-rs yesterday, it will be cool if running on web browsers in the future instead of by poor WebGL library.  I am planning to port some examples in [https://github.com/SaschaWillems/Vulkan](url) with wgpu-rs.

kvark:
Bors r=vitvakatu

PR #235, opened 2020-03-05
Update rustfmt configuration for current toolchains

ecton:
The goal of this PR is to configure `rustfmt` so that it can safely be run on this project, and bundle any associated formatting changes into one PR.

### DIFF
--- a/examples/anim.rs
+++ b/examples/anim.rs
@@ -87,7 +87,7 @@ fn make_indices(meta: &[u16]) -> Vec<[u32; 3]> {
 
 fn main() {
     let mut win = three::Window::new("Three-rs mesh blending example");
-    let cam = win.factory.perspective_camera(60.0, 1.0 .. 1000.0);
+    let cam = win.factory.perspective_camera(60.0, 1.0..1000.0);
     cam.look_at(
         [100.0, 0.0, 100.0],
         [0.0, 0.0, 30.0],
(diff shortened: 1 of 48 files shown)


ecton:
The output of the stable version of rustfmt after removing `max_width` was:

```
Warning: Unknown configuration option `fn_args_density`
Warning: Unknown configuration option `reorder_extern_crates`
Warning: Unknown configuration option `reorder_extern_crates_in_group`
Warning: Unknown configuration option `reorder_imported_names`
Warning: Unknown configuration option `reorder_imports_in_group`
Warning: can't set `spaces_around_ranges = true`, unstable features are only available in nightly channel.
Warning: can't set `error_on_line_overflow = false`, unstable features are only available in nightly channel.
```

`error_on_line_overflow` doesn't appear to affect this project, but `spaces_around_ranges` requires the nightly version of rustfmt. I removed the options that are unknown. Do you want to keep the requirement on using the nightly toolchain? 

The only reason I ask is because I'm using rust-analyzer in my daily development, and I can't seem to configure it to work with the [nightly toolchain (short of building it myself)](https://github.com/rust-analyzer/rust-analyzer/blob/ce663ed4954361a0421ff4b46747cb7c197e0204/xtask/src/lib.rs#L28). Still, I'm willing to run `cargo +nightly fmt` before pushing changes if you would prefer not to change that.

Let me know what your thoughts are!

kvark:
Thank you for the PR!
We should definitely be using the nightly rustfmt.

PR #236, opened 2020-05-13
Update obj

elrnv:
Update the obj dependency to 0.10 and integrate the changes.


kvark:
Actually, looks like gltf fails to compile now :/
https://travis-ci.org/github/three-rs/three/jobs/686387907#L874

kvark:
bors retry

